PR #72, opened 2020-08-18
✨ Allow relative URLs in bottom bar

### LINKED ISSUE #71: Allow relative URLs in bottom bar
Consider an index.gmi that exposes some subdirectories but not others (e.g a ./hidden link that's not in index.gmi).
(There are many such use-cases, such as leaving the door open to people with the link, or simply not wanting to have a directory exposed in the index, since most viewers will not care about it.)

In the current scenario, as far as I can tell, the only way to navigate to it is to type out the entire current prefix followed by /hidden, but the current path may be long.
It would be nice if one could type `./hidden` in the bottom bar, and (if the current url is a real one) the resolution happens internally.

I notice that amfora already has this for link handling (`followLink`), but `URL()` specifically wants absolute paths.

There's a couple of ways I can think of implementing this, but the calls are kind of all over the place.
Could this feature be added for convenience's sake?
I think a "good enough" heuristic should be if the first character of the bottom bar is "." - this way the method can also take care of handling ".." (but also further, things like "../otherdir").

CosmicToast:
Side effects:
If the URL does not end in /, any ./ action will actually operate on the parent directory.
For example, if the current path is /a/b/c, ./d will go to /a/b/d.
This is expected in the RFC (as far as I can tell), but it's a potential "gotcha" in case of naive servers & co.

Fixes #71

### DIFF
--- a/display/display.go
+++ b/display/display.go
@@ -126,7 +126,8 @@ func Init() {
 				reset()
 				return
 			}
-			if query[0] == '.' && tabs[tab].hasContent() { // relative url
+			if query[0] == '.' && tabs[tab].hasContent() {
+				// Relative url
 				current, err := url.Parse(tabs[tab].page.Url)
 				if err != nil {
 					// This shouldn't occur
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
Great, thanks! Merged.

>  For example, if the current path is /a/b/c, ./d will go to /a/b/d.

This is what I expect, all good with me.

PR #74, opened 2020-08-24
✨Add option to open unknown URLs

sotpapathe:
A new config option that sets a command to open unsupported URL schemes (e.g. `gopher://`). It is unset by default to keep the previous behavior.

### DIFF
--- a/default-config.toml
+++ b/default-config.toml
@@ -55,6 +55,20 @@ emoji_favicons = false
 shift_numbers = "!@#$%^&*()"
 
 
+[url-handlers]
+# Allows setting the commands to run for various URL schemes.
+# E.g. to open FTP URLs with FileZilla set the following key:
+#   ftp = "filezilla"
+# You can set any scheme to "off" to disable handling it.
+#
+# DO NOT use this for setting the HTTP command.
+# Use the http setting in the "a-general" section above
+
+# This is a special key that defines the handler for all URL schemes for which
+# no handler is defined.
+other = "off"
+
+
 [cache]
 # Options for page cache - which is only for text/gemini pages
 # Increase the cache size to speed up browsing at the expense of memory
--- a/display/private.go
+++ b/display/private.go
@@ -161,6 +161,30 @@ func handleHTTP(u string, showInfo bool) {
 	App.Draw()
 }
 
+// handleOther is used by handleURL.
+// It opens links other than Gemini and HTTP and displays Error modals.
+func handleOther(u string) {
+	// The URL should have a scheme due to a previous call to normalizeURL
+	parsed, _ := url.Parse(u)
+	// Search for a handler for the URL scheme
+	handler := strings.TrimSpace(viper.GetString("url-handlers." + parsed.Scheme))
+	if len(handler) == 0 {
+		handler = strings.TrimSpace(viper.GetString("url-handlers.other"))
+	}
+	switch handler {
+	case "", "off":
+		Error("URL Error", "Opening " + parsed.Scheme + " URLs is turned off.")
+	default:
+		// The config has a custom command to execute for URLs
+		fields := strings.Fields(handler)
+		err := exec.Command(fields[0], append(fields[1:], u)...).Start()
+		if err != nil {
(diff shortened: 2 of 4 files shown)


makeworld-the-better-one:
Thanks for this PR, it looks well done. My issue is that an unknown URL could be anything, and so having just one command for it isn't really portable. It could be any number of schemes: `gopher://`, `ipfs://`, `magnet:`, etc. Perhaps using `xdg-open` is the solution? I'm not sure, as that's often harder to configure.

sotpapathe:
It allows the user to at least specify another program that will handle unknown schemes, like `xdg-open` or [`pso`](https://github.com/galatolofederico/pso).

I guess the better approach would be to have a dictionary of scheme-program in the configuration, although I don't know how to implement that in go yet.

sotpapathe:
OK, I've got a better suggestion. The config file will look like this:

``` toml
[url_handlers]
http = "xdg-open"
gopher = "off"
ftp = "off"
# and any other URL schemes the user desires.
other = "xdg-open"
```

Then in `handleURL()`, if the scheme is not `gemini`, we'll do a `viper.GetString("url_handlers." + scheme)`. If a value is found, it's used as the program to open the URL with. Otherwise the value in `"url_handlers.other"` is used as the program to open the URL with.

If this sounds good I think I can implement it in the next days.

makeworld-the-better-one:
That sounds great! Good idea. I think we can leave the HTTP setting in `a-general` for compatibility though.

makeworld-the-better-one:
Make sure to update `default-config.toml` as well as `default.go`. I tend to edit the TOML first, then run `go generate ./...`.

sotpapathe:
Thanks for the tip, will do!

sotpapathe:
Updated as discussed and tested with Gopher URLs. I didn't know of any pages with other kinds of URLs but it should work for all schemes.

default-config.toml
```diff
@@ -55,6 +55,17 @@ emoji_favicons = false
 shift_numbers = "!@#$%^&*()"
 
 
+[url-handlers]
+# Allows setting the commands to run for various URL schemes.
+# E.g. to open FTP URLs with FileZilla set the following key:
+#   ftp = "filezilla"
+# You can set any scheme to "off" to disable handling it.
+
```
makeworld-the-better-one, line 63:
```suggestion

# DO NOT use this for setting the HTTP command.
# Use the http setting in the "a-general" section above.
```

Remember to add this change to `default.go` as well.

display/private.go
```diff
@@ -161,6 +161,31 @@ func handleHTTP(u string, showInfo bool) {
 	App.Draw()
 }
 
+// handleOther is used by handleURL.
+// It opens links other than Gemini and HTTP and displays Error modals.
+func handleOther(u string) {
+	// The URL should have a scheme due to a previous call to normalizeURL
+	parsed, _ := url.Parse(u)
+	scheme := parsed.Scheme
```
makeworld-the-better-one, line 169: This seems unnecessary since the variable is already available, it's not a function or anything.

default-config.toml
sotpapathe: Good point!

display/private.go
sotpapathe: I'm not sure whether I understood correctly. Do you suggest that the function prototype becomes

``` go
func handleOther(u URL)
```

makeworld-the-better-one: No, I'm saying line 169 is unnecessary, as you don't need a `scheme` variable. `parsed.Scheme` can be used everywhere instead. Not a big deal though.

sotpapathe: Oh, got it! Updated the code, it was a simple fix anyway.

makeworld-the-better-one:
Looks good now, thanks. I will merge soon, after my current linting fixes, like with the other PR. Also I'd appreciate if you didn't force-push for any future PRs, as it makes it impossible to see the diffs as you update things.

Thanks for your interest in Amfora!

sotpapathe:
Sorry for the force pushes, won't do it again :smiley: 

And thanks for making Amfora!

PR #75, opened 2020-08-24
✨Add option to automatically follow redirects

sotpapathe:
Should I also implement an upper limit on the maximum number of automatic redirects?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -144,6 +144,7 @@ func Init() error {
 	}
 
 	viper.SetDefault("a-general.home", "gemini.circumlunar.space")
+	viper.SetDefault("a-general.auto_redirect", false)
 	viper.SetDefault("a-general.http", "default")
 	viper.SetDefault("a-general.search", "gus.guru/search")
 	viper.SetDefault("a-general.color", true)
--- a/config/default.go
+++ b/config/default.go
@@ -16,6 +16,11 @@ var defaultConf = []byte(`# This is the default config file.
 # Press Ctrl-H to access it
 home = "gemini://gemini.circumlunar.space"
 
+# Follow up to 5 Gemini redirects without prompting.
+# A prompt is always shown after the 5th redirect and for redirects to protocols other than Gemini.
+# If set to false, a prompt will be shown before following redirects.
+auto_redirect = false
+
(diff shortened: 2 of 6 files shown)


makeworld-the-better-one:
Thanks for your interest and PR!

> Should I also implement an upper limit on the maximum number of automatic redirects?

Definitely. Five is recommended by web RFCs.

I'm not sure how good I feel about having the automatic redirects available in general, due to the potential for loops and cross-protocol redirects. If you want to write code handling those situations too, I would feel better about it. Otherwise it is a potentially dangerous feature.

sotpapathe:
So in short:

1. `auto_redirect` will allow up to 5 redirects without prompting the user.
2. Always prompt for redirects to protocols other than Gemini.

The prompt after 5 redirects will make it possible for the user to detect loops. For automatically detecting loops we'd have to keep a list of all previous URLs in the redirect series, since we can't know the period of the loop. But this way, if a malicious server creates a never-ending redirect sequence (e.g. `gemini://example.com/n` -> `gemini://example.com/n+1`), the client will run out of memory. Thus I suggest leaving loop detection to the user. Thoughts?

makeworld-the-better-one:
Both those things sound good, yep!

makeworld-the-better-one:
Make sure to update `default-config.toml` as well as `default.go`. I tend to edit the TOML first, then run `go generate ./...`.

sotpapathe:
Done! Tested using [gemini://gemini.conman.org/test/torture](gemini://gemini.conman.org/test/torture)

Not 100% satisfied with having to add another parameter to `handleURL()` but I don't know enough go yet to code a different solution. Suggestions welcome!

makeworld-the-better-one:
This looks good, thanks! [This line](https://github.com/makeworld-the-better-one/amfora/pull/75/files#diff-c14fd750fb214b4de03296a1c697fe04R424) is a clever solution.

I will merge to master soon, just have to finish up some linting fixes I'm doing there.

> Not 100% satisfied with having to add another parameter to `handleURL()` but I don't know enough go yet to code a different solution. Suggestions welcome!

I've had to do things like this for recursive functions before. One solution is to have another function, possibly internal, that uses the `numRedirects` argument, while the outer one like `handleURL` just sets it to `0`. But I'm fine with this for now, another functions would be annoying as well.

PR #76, opened 2020-08-27
Fix build error on darwin OS

sumpygump:
This fixes an error that prevented a successful build on darwin machines.

The error message was:

```
webbrowser/open_browser_unix.go:15:6: Open redeclared in this block
	previous declaration at webbrowser/open_browser_darwin.go:7:32
```

The issue was the build constraint in webbrowser/open_browser_unix.go
was being ignored because there was not a blank line following the build
constraint line.

I discovered the solution from here:
https://stackoverflow.com/questions/10646531/golang-conditional-compilation

### DIFF
--- a/webbrowser/open_browser_unix.go
+++ b/webbrowser/open_browser_unix.go
@@ -1,4 +1,5 @@
 // +build linux freebsd netbsd openbsd
+
 //nolint:goerr113
 package webbrowser
 


makeworld-the-better-one:
It appears I should stop doing things on master, as people are using it a lot. 

I'm aware of this issue, and have a fix for it on my computer, just didn't commit it yet because I was in the middle of fixing other things. Will commit it ASAP, and stop testing things out on master so much.

Thanks!

makeworld-the-better-one:
Fixed in 6853bb9. You can see that testing and building (on Linux) succeeded [here](https://travis-ci.com/github/makeworld-the-better-one/amfora/builds/181666108), where [previously](https://travis-ci.com/github/makeworld-the-better-one/amfora/builds/181402580) it failed.

PR #78, opened 2020-08-29
Fixes #55 - ignore navigation keypresses while help is open

### LINKED ISSUE #55: You can navigate through history when the help page is open

sumpygump:
The issues is that keypresses while the help is open are applied on the current document "underneath" the help table. This solution employs a similar technique that is used to ignore disallowed keypresses while a modal is open.

I also updated the help message to include help on page-up, page-down, u and d which were missing.

I also updated it to allow pressing enter to close the help too.

### DIFF
--- a/display/display.go
+++ b/display/display.go
@@ -229,6 +229,11 @@ func Init() {
 			// It's focused on a modal right now, nothing should interrupt
 			return event
 		}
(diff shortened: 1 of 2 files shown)


makeworld-the-better-one:
Thanks for fixing this! Merged.

Issue #67, opened 2020-08-08
Custom new tab page


Jackymancs4:
Hello @makeworld-the-better-one 
I see here:
https://github.com/makeworld-the-better-one/amfora/blob/master/display/newtab.go

That you hard coded the content of the new tab page. After a while I almost learned it by memory.

It would be cool if instead the page was loaded from e.g. `$HOME/.config/amfora/newtab.gmi` (or relative XDG setting). This would allow for higher level of customization, both static (a short list of uber-favourite list of sites), and dynamic (like a cronjob that pipe `fortune` messages into it).

p.s
The title is not very descriptive. Feel free to rename it.

makeworld-the-better-one:
This would be a cool feature, thanks. It'd be nice if the page could be cached on startup, but would that break your dynamic idea?

Jackymancs4:
> but would that break your dynamic idea?

No, not really. It feels like a potentially different newtab page per session (instead of per tab open) is more than enough for cool ideas to emerge.

Still, as potential fix, you could allow the reload function (`ctrl-r`) to also affect the newtab page.

makeworld-the-better-one:
Added by #83, merged into master at 9d7c7370e087cce1db3f9a24dfd39d5ddbc3d307. It will be in v1.5.0.

PR #83, opened 2020-08-31
Custom new tab page

sotpapathe:
The new tab page can now be loaded from `$XDG_CONFIG_HOME/amfora/new_tab.gmi`. This is documented in the default tab page. This implements what was suggested in issue #67 except for the ability to reload the new tab page, it is loaded once on startup.

- I didn't make the new tab file location configurable. It will require expanding `~` in the config option and possibly environment variables in order to honor `XDG_CONFIG_HOME` so I opted for a hardcoded path like the config file. If it is deemed necessary I can implement it.
- `config.ConfigDir` was exported to allow accessing `XDG_CONFIG_HOME` from outside `config`.
- I didn't implement the reload yet because it will require some special cases and I'm not sure where best to add them, `Reload()`, or `hasContent()` and `handleURL()`?

### DIFF
--- a/display/newtab.go
+++ b/display/newtab.go
@@ -1,4 +1,3 @@
-//nolint
 package display
 
 import (
@@ -7,6 +6,7 @@ import (
 	"github.com/makeworld-the-better-one/amfora/config"
 )
 
+//nolint
 var defaultNewTabContent = `# New Tab
 
 You've opened a new tab. Use the bar at the bottom to browse around. You can start typing in it by pressing the space key.
--- a/display/newtab.go
+++ b/display/newtab.go
@@ -30,7 +30,6 @@ func getNewTabContent() string {
 	data, err := ioutil.ReadFile(newTabFile)
 	if err == nil {
 		return string(data)
-	} else {
-		return defaultNewTabContent
 	}
+	return defaultNewTabContent
 }
--- a/display/newtab.go
+++ b/display/newtab.go
@@ -3,7 +3,6 @@ package display
 
 import (
 	"io/ioutil"
-	"path/filepath"
 
 	"github.com/makeworld-the-better-one/amfora/config"
 )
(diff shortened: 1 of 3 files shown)


display/newtab.go
```diff
@@ -1,16 +1,40 @@
...
 You've opened a new tab. Use the bar at the bottom to browse around. You can start typing in it by pressing the space key.
 
 Press the ? key at any time to bring up the help, and see other keybindings. Most are what you expect.
 
+You can customize this page by creating the Gemtext file
+
+` + "```" + `
+$XDG_CONFIG_HOME/amfora/new_tab.gmi
+` + "```" + `
+
```
makeworld-the-better-one, line 22:
```suggestion
You can customize this page by creating a gemtext file called newtab.gmi, in Amfora's configuration folder.
```

display/newtab.go
```diff
@@ -1,16 +1,40 @@
...
+// Read the new tab content from a file if it exists or fallback to a default page.
+func getNewTabContent() string {
+	newTabFile := filepath.Join(config.ConfigDir, "new_tab.gmi")
+	data, err := ioutil.ReadFile(newTabFile)
+	if err == nil {
+		return string(data)
+	} else {
+		return defaultNewTabContent
+	}
```
makeworld-the-better-one, line 39:
```suggestion
	}
	return defaultNewTabContent
```

sotpapathe:
Done! I didn't add the `CustomNewTab` boolean because if the custom new tab page is created after startup, the custom page won't appear after reloading. With the current implementation, at each reload, a custom new tab page is searched for.

makeworld-the-better-one:
Thanks! The problem with not having `CustomNewTab` is that if there is no new tab page, the the default new tab is still re-rendered every time you open a new tab, even if the terminal is the same size.

> if the custom new tab page is created after startup, the custom page won't appear after reloading

I'm okay with this, because of the performance gain.

sotpapathe:
Fair point, done. If a custom new tab page is present on startup, then the new tab page will always be reloaded. Otherwise it will never be reloaded, even if a custom new tab page appears later.

Should I squash my commits before merging?

makeworld-the-better-one:
Thanks for all your work! Merged. And no, you don't have to squash, I can do that with the "Squash and merge" button.

sotpapathe:
Got it. Thanks for merging and making amfora!

Issue #59, opened 2020-07-31
Only support ANSI codes in preformatted blocks


makeworld-the-better-one:

sumpygump:
I researched this and discovered what is going on. The gist is the ANSI codes are being reset for each line.

In renderer.go, all the highlight region definitions like `"[" + config.GetColorString("regular_text") + "]"` and `[-::-]` and `[-]` are embedded on every line, so if you set a color in the gmi content with an ANSI code, it will not keep the color to the next line(s). It works in the case where the ANSI code is on the same line as the gmi content. This is because the renderer will apply the `regular_text` color for example first and then the ANSI code in the line's content will be applied and then finally at the end of the line the renderer will reset the color. Even when a long line is wrapped, each wrapped line will be reset too.

The following snippet is an example that illustrates the above description. Note, of course to test it out on a real page, you must convert the `^[` to the escape character, but I have supplied a screenshot from Amfora below it:

```
^[[94m Setting the color on the same line works ^[[0m
^[[94m 
But trying to set the color from the previous line doesn't work because it gets reset.

^[[94m But what if it is a really long line like this one? When it renders the wrap, each new line gets a color reset. lorem ipsum lorem ipsum lorem ipsum lorem ipsum  lorem ippsum lorem ipsum lorem ipsum lorem ipsum  lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum lorem ipsum  lorem ipsum lorem ipsum lorem ipsum llorem ipsum  lorem ipsum lorem ipsum lorem ipsum lorem ipsum  lorem ipsum lorem ipsum lorem ipsum lorem ipsum  lorem ipsum lorem ipsum lorem ipsum lorem ipsum  lorem ipsum loreem ipsum lorem ipsum lorem ipsum.

Even a link could have some ANSI color codes in the middle of it (the word
middle is in yellow), but unfortunately a op() (0m) to reset the color changes
off the default for the link.

=> gemini://link What if a color code is the ^[[93mmiddle^[[0m of a link text?
```

![image](https://user-images.githubusercontent.com/676007/91628522-242c6480-e986-11ea-9534-2b3a4dcee1f0.png)

For the makeworld.gq banner, it could be fixed in the content by adding the `^[[94m` escape code at the beginning of each line of the ascii art. In general, I am not sure what a long-term solution for supporting ANSI color codes in page content would be.

[...]

PR #86, opened 2020-09-03
Strip ANSI codes from page based on config - fixes #79

### LINKED ISSUE #79: Strip ANSI codes from page
This should happen when:
- [ ] `color = false`
- [ ] `ansi = false` - New setting, doesn't exist yet

This can be done with a small regex, see [here](https://superuser.com/questions/380772/removing-ansi-color-codes-from-text-stream).

sumpygump:
Adds new boolean config option `ansi` in `[a-general]` to switch whether ANSI codes in page content will be rendered to terminal.

If `ansi `is set to false or `color` is set to false, any ANSI codes will be stripped from fetched content.
Fixes #79 

Here is a summary of results for the various values of the two options:

| config: color | config: ansi | Result                                    |
|---------------|--------------|-------------------------------------------|
| true          | true         | colorize headings & links; translate ANSI |
| true          | false        | colorize headings & links; strip ANSI     |
| false         | true         | no color; strip ANSI                      |
| false         | false        | no color; strip ANSI                      |

Notice how it is not possible to have `ansi` color codes true and `color` false at the same time. If the `color` config option is false, it overrides the value of `ansi`.

Feature to only allow ansi codes in preformatted blocks is part of another issue (#59)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -206,6 +206,7 @@ func Init() error {
 	viper.SetDefault("a-general.http", "default")
 	viper.SetDefault("a-general.search", "gus.guru/search")
 	viper.SetDefault("a-general.color", true)
+	viper.SetDefault("a-general.ansi", true)
 	viper.SetDefault("a-general.bullets", true)
 	viper.SetDefault("a-general.left_margin", 0.15)
 	viper.SetDefault("a-general.max_width", 100)
--- a/config/default.go
+++ b/config/default.go
@@ -33,6 +33,9 @@ search = "gemini://gus.guru/search"
 # Whether colors will be used in the terminal
 color = true
 
+# Whether ANSI codes from the page content should be rendered
+ansi = true
+
 # Whether to replace list asterisks with unicode bullets
 bullets = true
 
--- a/default-config.toml
+++ b/default-config.toml
@@ -30,6 +30,9 @@ search = "gemini://gus.guru/search"
 # Whether colors will be used in the terminal
 color = true
 
+# Whether ANSI codes from the page content should be rendered
+ansi = true
+
 # Whether to replace list asterisks with unicode bullets
 bullets = true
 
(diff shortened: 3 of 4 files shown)


makeworld-the-better-one (changes requested):
Thanks, this is helpful! I have a few issues.

- What is the source for that regex? You should put it as a comment in the code
- Define the regex as a package/global variable like `ansiRegex` or `ansiRE`
- Optionally strip ANSI in the `RenderANSI` function as well
- Add the default value of `a-general.ansi` to `config/config.go`, like all config options

You can add `//nolint:lll` right before the regex variable to make linting pass, I don't mind in this case.

sumpygump:
Thanks for the tips. I made your suggested changes.

makeworld-the-better-one:
The `ansiRegex` variable should be the output of `regexp.MustCompile`, not a string. Also I'd appreciate if you pointed to the source directly:
```go
// Source: https://github.com/acarl005/stripansi/blob/2749a054d4758fb0126d83aae590d7c9e8982d77/stripansi.go
```

On another level, I wonder if there's any point in stripping non-color codes only sometimes. The way things are done currently, those codes will be displayed on the screen when ANSI is enabled, but stripped when it's disabled. It might make sense to only strip color codes, since the move and other codes don't actually affect the terminal display anyway, thanks to cview.

A regex like ``` `\x1b\[[0-9;]*m` ``` (note it's a backtick string) should be able to just get the color codes. That's taken from [here](https://superuser.com/questions/380772/removing-ansi-color-codes-from-text-stream), which I linked in #79. I'm not sure why the stripansi regex is so long compared to these other ones. I've seen this longer version in other places like [here](https://github.com/chalk/ansi-regex/blob/master/index.js), but there seems to be no complaints about the shorter ones on superuser.com.

sumpygump:
Okay, that makes sense. I was thinking that the goal would be to strip all ansi codes, worrying that things like clear screen and move would potentially mess up the rendering of the page. I'll switch it to the simpler regex.

PR #87, opened 2020-09-05
Add 'e' keybinding feature to edit current URL

sumpygump:
One thing I found myself wanting to do from time to time is make a slight tweak to the URL I am currently viewing.

I added this feature to edit the URL of the current page. Press 'e' to have the bottomBar populate with the current URL in case you want to change the path or query param. It is a good way to go to a similar path or page based on the existing URL without having to type the whole thing out again.


makeworld-the-better-one:
Nice idea, thanks! Another way to do that would be with the relative URL feature added in the latest release: Enter a URL starting with a `.` to have it interpreted relative to the current one.

Issue #69, opened 2020-08-15
Emoji favicons on bookmarks page


makeworld-the-better-one:
Only if `emoji_favicons` is enabled.

```
# Bookmarks

[1] 🌻 Astrobotany

...etc...
```

alexwennerberg:
What do you think the best way to do this is? Set the emoji on write, or pull it when bookmarks page is loaded (cached)?

makeworld-the-better-one:
Added in #90.

PR #90, opened 2020-09-16
Emoji Favicons

alexwennerberg:
Hi! I decided I want to help contribute to this project when I have time, and I thought this would be a good issue to get started with -- #69

This isn't a perfect solution -- it just pulls from the cache right now if it's there. I can think of three solutions to this problem:

1. Persist the cache somehow on close / startup
2. Make gemini requests to grab favicons when opening bookmarks page, or on startup (seems slow, esp with many bookmarks)
3. Load icon only if it is already saved in the cache (my solution)

My solution is partway towards a solution based on 1. which seems to me like the best approach. What do you think?

Thanks for making this tool!

### DIFF
--- a/display/bookmarks.go
+++ b/display/bookmarks.go
@@ -2,13 +2,11 @@ package display
 
 import (
 	"fmt"
-	"net/url"
 	"strconv"
 	"strings"
 
 	"github.com/gdamore/tcell"
 	"github.com/makeworld-the-better-one/amfora/bookmarks"
-	"github.com/makeworld-the-better-one/amfora/cache"
 	"github.com/makeworld-the-better-one/amfora/config"
 	"github.com/makeworld-the-better-one/amfora/renderer"
 	"github.com/makeworld-the-better-one/amfora/structs"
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
Hello, thanks for contributing!

I didn't say this in the issue, I should have, but my preferred way of doing this is to actually write the favicon to the bookmarks file. Like when you press <kbd>Ctrl-D</kbd> to create a bookmark for the first time, the default text starts with the favicon for the page if it exists, and a space. And if the favicon for the page changes, you can delete the bookmark and make a new one, or change it yourself. This seems like a simpler solution that allows the user to control what the bookmarks look like too. This is basically your solution 1, "persist the cache".

> Load icon only if it is already saved in the cache

The main problem with this that I see is that since the cache is new on startup, usually most of the bookmarks won't have favicons at all. You'd have to visit all your bookmarked sites and never close Amfora for them to display the way you'd expect.

What do you think of my solution? Could you change your PR to implement it? Let me know.

alexwennerberg:
Yes, that makes sense! That was one solution I thought of but forgot to write it down in my list. I’ll implement the change after work today if i have time. It seems like the only disadvantage is if pages change their favicon or add one where there wasn’t one already but it also has the added advantage of users being able to delete/edit favicons for certain pages if they want

Alex

makeworld-the-better-one:
Ok, glad you agree! Thanks.

alexwennerberg:
OK, I think I have a change does what you're looking for -- let me know what you think!

makeworld-the-better-one:
Thanks for adding this! Looks mostly good, but one bug I noticed was that the favicon gets added twice if I try to change the bookmark. For example, if you visit a site with a favicon, and then create a bookmark, the default bookmark text is `<favicon> `. I can add a bookmark, like `<favicon> My Fav Capsule`, but if I go back to that site and try to change the bookmark (with <kbd>Ctrl-D</kbd>) again, the default text that comes up is not the actual bookmark text, but `<favicon> <favicon> My Fav Capsule`.

Maybe you could add code that checks if the bookmark already starts with the current site's favicon, and only prepend the favicon if it doesn't?

alexwennerberg:
Should be fixed now! However, I realized there seems to be a bug with updating bookmarks -- created a separate issue here: https://github.com/makeworld-the-better-one/amfora/issues/92

makeworld-the-better-one:
Looks like the original issue I had was fixed. Thanks for this PR!

alexwennerberg:
No problem! I’ll try and keep an eye on this repo and contribute when I have a chance

Alex

PR #94, opened 2020-10-01
Remove protocol filter for bookmarking

oholiab:
Apologies if I've missed some nuance that this breaks, but I figured it was more helpful to post a PR than an issue here.

This removes the protocol filter when adding a bookmark. I've been using amfora along with solderpunk's `agena` proxy (https://tildegit.org/solderpunk/agena) for accessing gopher, and it works beautifully well.

With this patch, adding and visiting bookmarks works just fine.

I'm happy to do further testing with my setup if necessary.

### DIFF
--- a/display/bookmarks.go
+++ b/display/bookmarks.go
@@ -3,7 +3,6 @@ package display
 import (
 	"fmt"
 	"strconv"
-	"strings"
 
 	"github.com/gdamore/tcell"
 	"github.com/makeworld-the-better-one/amfora/bookmarks"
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
Glad to hear Amfora is working well for you! Thanks for your contribution.

I did some testing and I don't think this will cause any issues. Merged.

oholiab:
Excellent! Thank you so much for Amfora, it's so satisfying to use and very beautiful :)

PR #98, opened 2020-10-02
Xdg user dirs

### LINKED ISSUE #97: Parse XDG config file for user dirs
[XDG user dirs](https://wiki.archlinux.org/index.php/XDG_user_directories) specify the user's download directory, so if this is set amfora should use these directories for saving downloaded documents.

pniedzwiedzinski:
I added user-dirs support for downloads dir. This will fix #97


makeworld-the-better-one:
Thank you for your contribution, but I don't think I can merge this. It would be more reliable to parse out the variable in Go instead of using `sh`, which will also fail everytime on platforms like Windows.

Also, if you would like to add a Nix pkg for Amfora, please do that in a separate PR.

I have added a comment in #97 about how I would accomplish this, if you would like to take another shot.

PR #101, opened 2020-10-11
Remove basedir check

pniedzwiedzinski:
Basedir checks are not needed as it is set by default. See: https://github.com/rkoesters/xdg/blob/master/basedir/basedir.go#L52-L66


makeworld-the-better-one:
Thanks!

PR #102, opened 2020-10-20
🐛 Allow closing any tab

sotpapathe:


makeworld-the-better-one:
Thanks a lot for adding this. I feel somewhat unsure about it right now, because although I did some basic testing and it works well, this is a prime spot for race conditions. If a tab is loading or an error pops up, and then the tab is closed, that sort of thing. I also feel unsure whether `tabPages` should be removed first, or if `tabs` should be altered first. Is there a reason behind the way you did it in your PR currently?

I would like to merge this, I just don't want to introduce difficult/complicated bugs. I want to feel sure about potential race conditions before merging.

sotpapathe:
> I also feel unsure whether tabPages should be removed first, or if tabs should be altered first. Is there a reason behind the way you did it in your PR currently?

I removed the `tabPages` first and the `tab` afterwards because it seemed like the most straightforward and cleanest implementation. I haven't though about potential race conditions so of course I would change the code if it causes any.

> If a tab is loading or an error pops up, and then the tab is closed, that sort of thing.

I haven't tested this particular case, putting it on my TODO list.

> I would like to merge this, I just don't want to introduce difficult/complicated bugs. I want to feel sure about potential race conditions before merging.

Absolutely understandable. I'll do some more testing and post my findings soon. Let me know if I can do anything else to make you feel more confident about merging this.

makeworld-the-better-one:
> > If a tab is loading or an error pops up, and then the tab is closed, that sort of thing.
>
> I haven't tested this particular case, putting it on my TODO list.

This is two cases:
- Tab is loading and then is closed
- A tab is closed, and at the same time an error pops up

They were just examples. The underlying idea is something involving a tab happening at the same time as that tab is being closed.

> I'll do some more testing and post my findings soon. Let me know if I can do anything else to make you feel more confident about merging this.

Thanks! Looking forward to seeing your results. As for me feeling more confident, I will have to look through the code and do a lot of thinking :brain::brain:. Race conditions are hard to find or verify experimentally, so I have to look through the code to see if this will cause one.

I am pretty busy at the moment, so I can't promise when this will happen, sorry. But I appreciate the PR and won't forget about it!

---

By the way, the ideal way to solve this would be some refactoring that protects all changes to `tabs` and `tabPages` variables with a mutex, so that nothing can happen in-between the change of one and the change of the other. Right now in the code, it's possible for code to run after `tabPages` has been modified but before `tabs` has been. That could cause issues, or not, but in any case it's not exactly the greatest code strategy.

makeworld-the-better-one:
The method through which tabs are closed has changed in #107, and so this PR might become irrelevant once that is merged. I will keep this open for now, since I am not sure yet if that PR will have any unforeseen issues.

makeworld-the-better-one:
Todo: Check for places that reference a tab by its index in `tabs` repeatedly, instead of using a pointer. Those places are where race conditions might occur, causing an action to be applied to the wrong tab.

sotpapathe:
I had a look at that a couple of days ago. It seems it would require quite a few mutex lock/unlock calls. Is this the way to go or is it easy to have a more clever design that automatically locks/unlocks on access? I'm not very well versed in GO yet...

makeworld-the-better-one:
Mutexes would be the way to go. If it makes sense, some funcs could be written than handle it "automatically", but there's nothing built-in for that that applies this case.

My todo comment was for myself, but I was going to look to see if I can get away without using mutexes. But I think you're right, and it would be a mistake to not use them.

In that case, the next steps for this PR would be to sync it with master (merge master into it), and then create a `RWMutex` that's used every time `tabs` is accessed. If you want to take a stab at that, let me know. Otherwise I will do it at some point. No worries either way.

sotpapathe:
I might work on it during the coming holidays. I'll make sure to check if you've created a relevant branch and also post here before starting work on this.

makeworld-the-better-one:
Alright, thanks. Please also let me know if you decide you won't be or can't work on this. And I'll let you know here if I start working on it instead, although I expect I'll be focusing on other things.

makeworld-the-better-one:
This is too outdated now, sorry it never got merged. Thanks for you work nonetheless.

sotpapathe:
No worries! Sorry I didn't get around to finishing it.

PR #104, opened 2020-10-26
Update dependencies, browse using TabbedPanels

tslocum:
Favicon support must be re-implemented.


makeworld-the-better-one:
Wow, having the upstream maintainer come and update your project for you is really phenomenal. Thank you, this is very much appreciated. This was going to be a lot of tedious text work for me, and it's great to have it complete. I also appreciate the simplification of the help page code, it seems obvious now that `text/tabwriter` should have been used.

To re-add favicon support as you mentioned, this is correct function, right?
```go
func (t *TabbedPanels) SetTabLabel(name, label string)
```

So that the label of the tab is changed to the emoji.

makeworld-the-better-one:
I have changed the base branch to `cview-update`, which is where I had started updating cview previously. I will re-implement favicons there and do lots of testing before merging to master. If you realize you missed something or weren't finished with this PR, feel free to add a comment in #107 or make another PR. Thanks again!

makeworld-the-better-one:
Also, one thing I think you missed is adding `.Truecolor()` when loading custom colors from the config. Because those colors could be any RGB colors, not just from the xterm 256. I will add that in #107, let me know if I'm misunderstanding.

tslocum:
Yes, that is the correct function to use, and I did miss updating to use `TrueColor` values.  Glad to help.

Issue #61, opened 2020-08-03
Subscriptions


makeworld-the-better-one:
- [x] A dynamically constructed (but possibly cached) feed viewer page, similar to the way CAPCOM and Spacewalk look
- [x] Automatic detection of feed files
	- Filenames: `atom.xml`, `feed.xml`, `*.rss`, `*.atom`, `feed.json`
	- Mimetypes: `application/atom+xml`, `application/rss+xml`, `application/json+feed`
	- [x] User will be prompted to add the feed
- [x] Some shortcut or button to add a URL at any time
	- If it's a valid feed it will be added as one, otherwise the page will be tracked
- [x] A way to edit the feed list and remove feeds
	- Maybe by visiting the page or any post URL and doing a delete shortcut?
- [x] Pagination

When should the feed view be regenerated?

- ~~On hard-reload of the feed viewer~~
- ~~On every X load of the feed viewer~~
- [x] At browser start
- [x] Every X minutes in a separate thread (30 mins by default)

makeworld-the-better-one:
[gofeed](https://github.com/mmcdole/gofeed) for RSS and Atom feeds.

makeworld-the-better-one:
The name for this needs to be unified before merging. Feeds are "subscribed" to, pages are "tracked", but also feeds are "tracked", etc. The name "feeds" is used throughout the code, but this is not representative due to the page tracking functionality.

Maybe just "Subscriptions"?
- subscribe to this feed
- subscribe to this page
- remove this subscription / unsubscribe
- refresh this subscription

makeworld-the-better-one:
The "gemini feed format" (`gemini://gemini.circumlunar.space/docs/companion/subscription.gmi`) will not be supported initially. A separate issue will be opened, and it will likely be in a later release.

makeworld-the-better-one:
Sometimes corrupted JSON files are being written. This is critical error that must be discovered and fixed before merging into master. Here is an example of the bottom of a corrupted file:

```json
    },
    "gemini://republic.circumlunar.space/users/xkp/gemlog/": {
      "hash": "24860ed36333d360817b5d35614fdd714266df360c54e35a2f1ac1ce7375d82a",
      "changed": "2020-11-24T18:34:49.418540845Z"
    }
  }
}
r.space/users/xkp/gemlog/": {
      "hash": "24860ed36333d360817b5d35614fdd714266df360c54e35a2f1ac1ce7375d82a",
[...]

PR #112, opened 2020-11-04
Allow specifying a client certificate

tslocum:


makeworld-the-better-one:
Thanks for making this PR. However, as I've mentioned on the README, I would like to add advanced client cert support, where multiple certs are supported for different sites, and can be generated within the client.

> - [ ] Full client certificate UX within the client
>	- Create transient and permanent certs within the client, per domain
>	- Manage and browse them
>	- Similar to [Kristall](https://github.com/MasterQ32/kristall)
>	- https://lists.orbitalfox.eu/archives/gemini/2020/001400.html

I think adding this smaller support would be nice for some circumstances, but ultimately lead to issues as most people use different certs for different domains, as they should.

tslocum:
We could update the PR to point to a directory of per-domain certificates in the mean time.  Waiting for full in-client UX also sounds good.

makeworld-the-better-one:
Maybe I shouldn't have closed so prematurely. I think adding multi-cert support at a lower level is a good idea, and the UI can come at a later date, when I have time.

I figure this will mostly be a configuration thing, we need a way to specify cert and key files per host (domain and port) in the config file. I don't want to rely on a directory or assumed name schemes (like it will always be named `thedomain.com.crt`) so that certs can be shared across clients.

What do you think of this method?

```toml
[auth]

[auth.certs]
"example.com" = "mycert.crt"

[auth.keys]
"example.com" = "mycert.key"
```

Just off the top of my head. The viper keys would then be `"auth.certs." + host` and `"auth.keys." + host`.

tslocum:
Sure.  I've updated the PR.

makeworld-the-better-one (changes requested):
Please see the comments below. One other thing that should be added is a public function (in `client.go`):
```go
func HasClientCert(host string) bool
```

This would return true when there is a valid client cert available for the provided host. I think using the cache only is fine here, it might be slow to attempt to read a few bytes of a file when it's not in the cache.

This doesn't have to be part of this PR, but the reason why I want that func is to disable caching for domains that client certs are being used for. In my early testing of this feature, I found that interactive apps like Astrobotany used redirects to pages to display new content, and it wasn't clear that things had changed/updated, because the cached version was being displayed.

tslocum:
OK, I've updated the PR.

makeworld-the-better-one:
I would appreciate if you didn't force-push PRs in general, so I can see the diff between commits. Will take a look.

makeworld-the-better-one:
Thanks a lot for this!

tslocum:
You're welcome.  You can review changes between pushes by clicking on the text **force-pushed**.

makeworld-the-better-one:
I'd just like to let you know that I made some changes to this code on the `feeds` branch, in e026eb2. You can track the progress of the feeds feature overall in #61.

It wasn't a problem when you made this PR, but with the feeds feature there are now goroutines fetching things concurrently, which means that `certCache` needs to be protected with a mutex. I would occasionally get `concurrent map writes` panic.

The other change was that from what I can tell, the old code would check every single time for a cert if one was specified for the host but there were errors opening it. Now the code will ignore that cert after the first time getting errors opening it.

PR #113, opened 2020-11-05
Do not cache pages accessed with a client certificate

tslocum:


makeworld-the-better-one:
Thanks, I just added this in 4fbaa97 actually. Also, using `parsed.Host` is correct. Client certs may be different for different ports, which using `pared.Hostname()` would ignore.

PR #116, opened 2020-11-08
Allow using terminal default background as color

jedthehumanoid:
Sometimes this is desirable. For example using terminal default background (or transparent terminal) instead of filling with black comes to mind.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -230,7 +230,7 @@ func Init() error {
 				return fmt.Errorf(`value for "%s" is not a string: %v`, k, v)
 			}
 			color := tcell.GetColor(strings.ToLower(colorStr))
-			if color == tcell.ColorDefault && colorStr != "default" {
+			if color == tcell.ColorDefault && !(colorStr == "defaultbg" && k == "bg") {
 				return fmt.Errorf(`invalid color format for "%s": %s`, k, colorStr)
 
 			}
--- a/config/config.go
+++ b/config/config.go
@@ -230,8 +230,9 @@ func Init() error {
 				return fmt.Errorf(`value for "%s" is not a string: %v`, k, v)
 			}
 			color := tcell.GetColor(strings.ToLower(colorStr))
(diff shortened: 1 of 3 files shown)


makeworld-the-better-one:
Thanks for adding this. I think it would be helpful to document this in the config: `default-config.toml` and `config/default.go`.

However, in my testing this was only really useful for the background, but not things like text. I tested on a white terminal, and it correctly chose a white background, but when I set the `regular_text`  to `default`, the text went white too, making it unreadable. Have you had a different experience? It seems like this is mostly useful if you are switching between terminal colors a lot, on the same machine.

jedthehumanoid:
Thanks for taking the time to answer!

Yeah, I only tested with combinations with white background color, and made assumptions I should have.

I see what you mean, the problem is in renderer/renderer.go:256, where it uses getColorString() to get a cview-compatible color tag.

There are actually two problems, older tcell versions returns #FFFFFF for unset colors in tcell.Color.Hex(), this is fixed now.
But even working around this, cview (atleast the version used here) doesn't respect the tag [-] as colorDefault, but uses white. And I cant understand why, it looks fine.

Anyway, this was not the simple one-liner that I imagined, I will try to sort it out.

Background color works fine as that is set by tcells regular workflow, this was originally my use case, but I feel like foreground should be consistent with this as well.

jedthehumanoid:
Since setting of foreground color got way more complex than I wanted, I will drop investigation of that.

I adapted the naive solution for just supporting background. 
This was my use case anyway, and probably most interesting for others.

@makeworld-the-better-one: If you still want that you can merge it, otherwise you can just drop this PR which is fine by me.

makeworld-the-better-one:
I think I will close this PR, since just being able to set the background automatically but not the text isn't super useful. I think the majority of people will be better served by doing it manually in the config. Thank you for your work, and I'm sorry it turned out to be more complicated!

PR #117, opened 2020-11-08
fix for #103: Simple handling of local (file://) urls

### LINKED ISSUE #103: Open Local Gemini Files
Opening a `.gmi` or `.gemini` file that is stored locally does not seem to be implemented. It would be a nice feature to be able to view a document as you write it.

jedthehumanoid:
Simple implementation for rendering local (file://) documents.

When opening a `file://` url:
**If it is a directory:**
Render a page with links to all dirs and files in directory.
Links are made to all kinds of files, checking if reasonable to open is done when actually opening them (with error popups)

**If it is a file:**
- If not a gemini file, use mime.TypeByExtension to check if text/* otherwise bail
- render .gemini/.gmi as gemini
- else render as plaintext

# TODO
- [x] directory listing
- [x] Files
   - [x] prohibit binaries
   - [x] gemini files
   - [x] other plain text files
- [x] links to other local resources
- [x] links to non-file://urls works as usual
- [x] Test on Linux
- [x] Test on Windows

fixes #103

### DIFF
--- a/display/file.go
+++ b/display/file.go
@@ -76,6 +76,7 @@ func handleFile(u string) (*structs.Page, bool) {
 				Raw:       content,
 				Content:   rendered,
 				Links:     links,
+				Width:     termW,
 			}
 		} else {
 			page = &structs.Page{
@@ -84,6 +85,7 @@ func handleFile(u string) (*structs.Page, bool) {
 				Raw:       content,
 				Content:   renderer.RenderPlainText(content, leftMargin()),
 				Links:     []string{},
+				Width:     termW,
 			}
 		}
 	}
@@ -117,6 +119,7 @@ func createDirectoryListing(u string) (*structs.Page, bool) {
 		Raw:       content,
 		Content:   rendered,
 		Links:     links,
+		Width:     termW,
 	}
 	return page, true
 }
--- a/display/file.go
+++ b/display/file.go
@@ -2,12 +2,11 @@ package display
 
 import (
 	"fmt"
-	"io"
 	"io/ioutil"
+	"mime"
 	"os"
 	"path/filepath"
 	"strings"
-	"unicode/utf8"
 
 	"github.com/makeworld-the-better-one/amfora/renderer"
 	"github.com/makeworld-the-better-one/amfora/structs"
@@ -35,45 +34,29 @@ func handleFile(u string) (*structs.Page, bool) {
 			return page, false
 		}
 
-		file, err := os.Open(filename)
+		mimetype := mime.TypeByExtension(filepath.Ext(filename))
+		if strings.HasSuffix(u, ".gmi") || strings.HasSuffix(u, ".gemini") {
+			mimetype = "text/gemini"
+		}
+
+		if !strings.HasPrefix(mimetype, "text/") {
+			Error("File Error", "Cannot open file, unknown mimetype.")
+			return page, false
+		}
+
+		content, err := ioutil.ReadFile(filename)
+
 		if err != nil {
 			Error("File Error", "Cannot open local file: "+err.Error())
 			return page, false
 		}
-		defer file.Close()
 
-		// Read first bytes, to check if plaintext
-		buf := make([]byte, 32)
-		_, err = file.Read(buf)
-		if err != nil && err != io.EOF {
-			Error("File Error", "Error reading file: "+err.Error())
-			return page, false
-		}
-
-		if !utf8.Valid(buf) {
-			Error("File Error", "Cannot open local file, looks like a binary.")
-			return page, false
-		}
-
-		// Looks like plaintext, keep reading
-		content := string(buf)
-		for {
-			_, err := file.Read(buf)
-			if err != nil {
-				if err != io.EOF {
-					Error("File Error", "Cannot open local file: "+err.Error())
-					return page, false
-				}
-				break
-			}
-			content += string(buf)
-		}
-		if strings.HasSuffix(u, ".gmi") || strings.HasSuffix(u, ".gemini") {
-			rendered, links := renderer.RenderGemini(content, textWidth(), leftMargin(), false)
+		if mimetype == "text/gemini" {
+			rendered, links := renderer.RenderGemini(string(content), textWidth(), leftMargin(), false)
 			page = &structs.Page{
 				Mediatype: structs.TextGemini,
 				URL:       u,
-				Raw:       content,
+				Raw:       string(content),
 				Content:   rendered,
 				Links:     links,
 				Width:     termW,
@@ -82,8 +65,8 @@ func handleFile(u string) (*structs.Page, bool) {
 			page = &structs.Page{
 				Mediatype: structs.TextPlain,
 				URL:       u,
-				Raw:       content,
-				Content:   renderer.RenderPlainText(content, leftMargin()),
+				Raw:       string(content),
+				Content:   renderer.RenderPlainText(string(content), leftMargin()),
 				Links:     []string{},
 				Width:     termW,
 			}
--- a/display/file.go
+++ b/display/file.go
@@ -5,9 +5,9 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 	"unicode/utf8"
-	"path/filepath"
 
 	"github.com/makeworld-the-better-one/amfora/renderer"
 	"github.com/makeworld-the-better-one/amfora/structs"
--- a/display/file.go
+++ b/display/file.go
@@ -16,7 +16,7 @@ import (
 func pathFromURI(u string) string {
 	path := strings.TrimPrefix(u, "file://")
 	path = strings.TrimPrefix(path, "localhost") // localhost is a valid host for file URIs
-	path = strings.TrimPrefix(path, "/") // Valid file URIs contains this additional slash
+	path = strings.TrimPrefix(path, "/")         // Valid file URIs contains this additional slash
 	return filepath.FromSlash(path)
 }
 
--- a/display/file.go
+++ b/display/file.go
@@ -22,7 +22,7 @@ func handleFile(u string) (*structs.Page, bool) {
 
 	fi, err := os.Stat(filename)
 	if err != nil {
-		Error("Cannot open local file", err.Error())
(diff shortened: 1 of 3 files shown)


makeworld-the-better-one:
Thanks for your work!

I see this PR is marked as a draft, so I won't look at the code yet. But to address your questions:

> Showing .txt files as well is straigtforward, but in that case we could show all sensible plain-text files regardless of extension, with some simple heuristic to ignore binary files and files too large.

That sounds good to me, I would like plain text support to be added. If that heuristic you mention ends up being too complicated or intensive, that just looking for *.txt is good with me too.

> Does not show directory listings for now. AFAICT normal gemini:// resouces does not do that either, so I am unsure about a idiomatic solution for doing this. Render directories as links and parsing it as a gemini-document?

Normal gemini sites aren't required to show directory listings, but most servers will if an `index.gmi/gemini` file doesn't exist. I would prefer it as it makes finding the file you want easier. And rendering directories as links is the way to go, yes. And I was picturing that only the files Amfora could open would be listed, so no `.jpg` or anything.

> Links to online resources works fine, but links to local files is wonky right now.

Linking to relative files can be fixed by creating a `handleFile` func, that contains most of the code you've added. Then in `handleURL`, make use of `handleFile` for two situations:

- When the URL starts with `file://`
- When the page/tab URL starts with `file://` and it's a local link

This should fix linking to local files.

jedthehumanoid:
Thanks for taking the time to investigate!

Yeah I wanted to ventilate alternatives before continuing.

Will give it a shot as discussed
- Show all plain-text files if not to complicated to evaluate if truly plain-text
- Render and show directory listing
- Fix local linking

makeworld-the-better-one:
Thanks! Please mark this as not-WIP when you're done and make a comment so I get notified.

jedthehumanoid:
@makeworld-the-better-one ok, tested on Windows as well as Linux

makeworld-the-better-one (changes requested):
Thanks a lot for adding this! It's a great feature to have. Please see my comments below.

display/file.go
```diff
@@ -0,0 +1,134 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/makeworld-the-better-one/amfora/renderer"
+	"github.com/makeworld-the-better-one/amfora/structs"
+)
+
+const maxSize = 1 * 1024 * 1024 // 1 Mb
```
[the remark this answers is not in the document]
jedthehumanoid: Naturally!, missed that one

display/file.go
```diff
@@ -0,0 +1,134 @@
...
+		Raw:       content,
+		Content:   rendered,
+		Links:     links,
+	}
+	return page, true
+}
+
+// resolveRelFileLink constructs a relative file:// link by matching it against
+// previous url
+func resolveRelFileLink(t *tab, prev, next string) string {
```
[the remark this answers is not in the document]
jedthehumanoid: Yeah, you are right! 
I added it because it didn't resolve right with the existing resolveRelLink, but then later i had to add a slash to dirs to make my resolver work correctly, looks like that's all what's needed to make existing resolver work as well. :weary:

display/file.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/makeworld-the-better-one/amfora/structs"
+)
+
+const maxSize = 1 * 1024 * 1024 // 1 Mb
+
+// handleFile handles urls using file:// protocol
+func handleFile(u string) (*structs.Page, bool) {
+	page := &structs.Page{}
+
+	filename := filepath.FromSlash(strings.TrimPrefix(u, "file://"))
```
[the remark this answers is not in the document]
jedthehumanoid: In this context, filename is used as a filepath (for os.open, read, readdir, ...).
I don't see how file URIs are relevant here, see FromSlash is run after file:// is removed from string.

I sure agree on renaming variable to path instead though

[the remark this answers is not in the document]
makeworld-the-better-one: You're right, I misunderstood. A valid file URI on Windows might be `file:///c:/WINDOWS/clock.avi`, but using `filepath.FromSlash(strings.TrimPrefix(u, "file://"))` on it will transform it into `\c:\WINDOWS\clock.avi`. I would like to be sure that Go will work properly with paths like that on Windows, seeing as it starts with a `\` and has a lowercase drive letter, but overall that function should stay in.

display/file.go
```diff
@@ -0,0 +1,134 @@
...
+
+		// Read first bytes, to check if plaintext
+		buf := make([]byte, 32)
+		_, err = file.Read(buf)
+		if err != nil && err != io.EOF {
+			Error("Error reading file", err.Error())
+			return page, false
+		}
+
+		if !utf8.Valid(buf) {
```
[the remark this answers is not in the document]
jedthehumanoid: Hmmm, yeah I see what you mean. 

How about **just** using mime.TypeByExtension for now and display those with "text/*", I think all the ones I imagined displaying are in that list anyway.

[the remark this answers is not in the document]
makeworld-the-better-one: Yeah that's probably fine. I already spotted some that one missed (`.py` for example), but that list is endless. Using just that is fine.

display/file.go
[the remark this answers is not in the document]
makeworld-the-better-one: > I would like to be sure that Go will work properly with paths like that on Windows

Please test this if you can and let me know the result.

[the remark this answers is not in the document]
jedthehumanoid: Yes it works now, did some additional checks after reading https://en.wikipedia.org/wiki/File_URI_scheme
Works fine on windows however I format it

jedthehumanoid:
Yes, also works with utf8 filenames and directories, just tested with my national åäö's, and also various wonky faces and symbols :sunglasses:

display/file.go
```diff
@@ -0,0 +1,113 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/makeworld-the-better-one/amfora/renderer"
+	"github.com/makeworld-the-better-one/amfora/structs"
+	"github.com/spf13/viper"
+)
+
+func pathFromURI(u string) string {
```
makeworld-the-better-one, line 16:
```suggestion
func pathFromFileURI(u string) string {
```

This func is available to all of `display/` so changing the name here makes it more clear.

display/file.go
```diff
@@ -0,0 +1,113 @@
...
+	path := strings.TrimPrefix(u, "file://")
+	path = strings.TrimPrefix(path, "localhost") // localhost is a valid host for file URIs
+	path = strings.TrimPrefix(path, "/")         // Valid file URIs contains this additional slash
```
makeworld-the-better-one, line 19: This additional represents the top level directory, does it not? For example, a file URI could be `file:///usr/local/bin/foo`. If you remove the "additional slash", you will end up with a relative path instead of an absolute one: `usr/local/bin/foo`.

A simple test for this is to make sure running `amfora file:///` opens the top level directory successfully. The next test is to make sure things work okay on Windows without it. If they don't, then I'm fine with removing that slash only on windows, by checking `runtime.GOOS`.

display/file.go
```diff
@@ -0,0 +1,113 @@
...
+			return page, false
+		}
+
+		mimetype := mime.TypeByExtension(filepath.Ext(path))
+		if strings.HasSuffix(u, ".gmi") || strings.HasSuffix(u, ".gemini") {
+			mimetype = "text/gemini"
+		}
+
+		if !strings.HasPrefix(mimetype, "text/") {
+			Error("File Error", "Cannot open file, unknown mimetype.")
```
makeworld-the-better-one, line 48:
```suggestion
			Error("File Error", "Cannot open file, not recognized as text.")
```

display/private.go
```diff
@@ -43,7 +43,9 @@ func followLink(t *tab, prev, next string) {
 
 	if t.hasContent() {
 		t.saveScroll() // Likely called later on, it's here just in case
-		nextURL, err := resolveRelLink(t, prev, next)
+		var nextURL string
+		var err error
+		nextURL, err = resolveRelLink(t, prev, next)
```
makeworld-the-better-one, line 48:
```suggestion
		nextURL, err := resolveRelLink(t, prev, next)
```

There's no need to add these lines.

makeworld-the-better-one (changes requested):
Thanks for updating the PR! It's almost ready, there's just one bug left, and some cosmetic changes.

display/private.go
jedthehumanoid: oops.

display/file.go
makeworld-the-better-one: Once this is fixed and tested on Windows, I should be good to merge.

jedthehumanoid: Hmmm, I had some problems thinking about a solution that solves relative (to current working directory) paths with file URIs.
But apparently that is not supported in any form in a valid file URI

Also looking at Go's url.ParseRequestURI which handles your comments and all other edge cases, but don't allow relative URIs.

Would you be ok with a solution that just disallow this (and follows the standard)?
In that case, just swithcing to url.ParseRequestURI is probably the way to go

makeworld-the-better-one: > Would you be ok with a solution that just disallow this (and follows the standard)?

Yep! This is how I expected it to work.

> `url.ParseRequestURI`

Sounds good to me, better than what I was suggesting. Just return `.Path` after parsing. The only extra thing would be to handle the possibility of a parsing error, ie a malformed URL. Maybe that would happen in this func, maybe somewhere else, which would probably remove the need for this func entirely. Make sure `Error` is called so the user is aware.

And don't forget to test on Windows. Thanks!

jedthehumanoid: @makeworld-the-better-one: Ok, tested on windows as well
Seems to work as expected, eventhough I'm not a fan of how Windows paths works in general.

https://github.com/golang/go/issues/6027 has a discussion on how Go chose to implement this.
In short, there is no perfect way to deal with windows drive letters, but go/python/nodejs/java have settled on one of the solutions.

makeworld-the-better-one: Thanks for the link and testing! The Windows stuff is a bit weird, but as long as the file opens correctly.

PR #118, opened 2020-11-13
Add the Nord theme

lokesh-krishna:
This is my theme, based on the [Nord](https://nordtheme.com) color scheme.  

Here is what it looks like:

### DIFF
--- a/contrib/themes/nord.toml
+++ b/contrib/themes/nord.toml
@@ -1,0 +1,111 @@
+[theme]
+# This section is for changing the COLORS used in Amfora.
+# These colors only apply if 'color' is enabled above.
+# Colors can be set using a W3C color name, or a hex value such as "#ffffff".
+
+# Note that not all colors will work on terminals that do not have truecolor support.
+# If you want to stick to the standard 16 or 256 colors, you can get
+# a list of those here: https://jonasjacek.github.io/colors/
+# DO NOT use the names from that site, just the hex codes.
+
+# Definitions:
+#   bg = background
+#   fg = foreground
+#   dl = download
+#   btn = button
+#   hdg = heading
+#   bkmk = bookmark
+#   modal = a popup window/box in the middle of the screen
+
+# EXAMPLES:
+# hdg_1 = "green"
+# hdg_2 = "#5f0000"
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
Looks great! Thanks for adding.

A sidenote: Why is the first tab symbol not a `1` in your screenshot, but that long dash character?

lokesh-krishna:
Thanks for merging :)  
I'm actually not sure about that character myself. I thought it was the default behaviour. What information would you need to see what's going on?

makeworld-the-better-one:
That's very strange. The relevant code is here:

https://github.com/makeworld-the-better-one/amfora/blob/8a1a845ab2755e342cdc7f402be0ec0ea9a7ad6a/display/private.go#L568-L583

The `char` variable holds the text that will be displayed up top. It should only ever be the tab number, or an emoji favicon. But that dash is not a valid emoji.

Here is what I get when I visit cosmic.voyage with `emoji_favicons` enabled:

```
  🚀  |  2  |                                                                                                                                                                                 
                                                                                                                                                                                              
                                      _. _  __._ _ * _.  .  , _   . _. _  _                                                                                                                   
                                     (_.(_)_) [ | )|(_.   \/ (_)\_|(_](_](/,                                                                                                                  
                                                                ._|   ._|                                                                                                                     
                                                                                                                                                                                              
                            Humanity has expanded to the stars. Earth is a distant memory.                                                                                                    
                                                                                                                                                                                              
                            We journey into the darkness not in a great migration--as many                                                                                                    
                            thought--but in small communities, drifting far and traveling at
```

With favicons disabled I just get a regular `1` as I should.

Is it possible you have `emoji_favicons` enabled in your config, but your terminal is replacing the emoji with that weird dash? That seems improbable to me but it's all I got.

I would also appreciate if you could run `amfora -v` for me and send the output, to see what we're working with.

lokesh-krishna:
I do have `emoji_favicons` enabled but I learnt only now that I didn't have an emoji font on my system. Installing JoyPixels fixed it for me :)

makeworld-the-better-one:
Great!

PR #133, opened 2020-11-30
show links

p1gp1g:


makeworld-the-better-one:
I'm not sure what this PR adds. It seems like link lines would become quite long like this. Personally if want to view the URL, I use the Tab and Shift-Tab keys. You can use Esc to stop selecting. What do you think?

p1gp1g:
Actually, I really don't like to clic on a link I haven't seen. The miss of this is even blocking for me. And I prefer to navigate with \<space\>+N°. For the moment I haven't seen a link written on 2 lines. 

I knew it would not be the choice of every body, that's why I set the default to false.

makeworld-the-better-one:
Alright. Please add the config changes to `default-config.toml` as well, and change the line that reads `linkText = linkText + " (" + url + ")"` to `linkText += " (" + url + ")"` instead. Thanks!

makeworld-the-better-one:
Thanks, merged.

p1gp1g:
Thanks :)

Issue #121, opened 2020-11-17
Mediatype handlers


makeworld-the-better-one:
Currently there is support for using different applications per URL scheme, see the `[url-handlers]` section of the config. Another section should be made to support defining a list of mediatypes that the specified application would handle. Currently the only option for binary files (images, audio, etc) is to download it or open it in the portal. Adding this would allow more seamless browsing.

The download modal buttons should change to "Open" and "Download". The "Open in Portal" button has always been a bit of a hack.

Example config, using an array of tables

```toml
# Explanation here...

[[mediatype-handlers]]
cmd = ["feh"]
types = ["image/jpeg"]


[[mediatype-handlers]]
cmd = ["vlc", "--flag"]
types = ["audio", "video"] # Can capture an entire type by just not specifying the subtype

[[mediatype-handlers]]
cmd = ["some-command"]
types = [
	"application/pdf",
	"*", # Override the default catch-all handler
]
```

- [x] Put this on the roadmap (also talk about scheme handlers in the same bullet maybe)
- [x] Add way to specify piping/streaming content versus downloading the entire thing first

makeworld-the-better-one:
On Unix systems default to using `xdg-open` for files if no other handler is found. Display an error to the user if there are no handlers and `xdg-open` does not exist.

For Windows and macOS do the same, but use the commands used in `webbrowser/` instead.

For other OSes just raise the error if there's no handlers.

makeworld-the-better-one:
The download modal would still display until the file is entirely downloaded and the application is launched -- except for piped/streamed mediatypes.

makeworld-the-better-one:
The `[url-handlers]` section of the config has this in it:

```toml
# This is a special key that defines the handler for all URL schemes for which
# no handler is defined.
other = 'off'
```

Maybe when this feature is added, then that key can be removed and ignored. And instead of displaying an error, when the user tries to open an unknown scheme, Amfora will fallback to the default opening command for the user's OS, as mentioned [above](https://github.com/makeworld-the-better-one/amfora/issues/121#issuecomment-732252868).

makeworld-the-better-one:
This feature has been added in #134 thanks to @sudobash1, in `mediatype-handlers` branch. Once the streaming part is added on that branch and v1.7.0 is released, then that branch can be merged to `master` and this issue can be closed.

makeworld-the-better-one:
[...]

Issue #129, opened 2020-11-25
Way to cancel downloading


makew0rld:
There should be a way to cancel downloading without using <kbd>Ctrl-C</kbd> and quitting the entire browser. Could easily be implemented by removing "Downloading..." in favour of a "Cancel" one that would close the connection and delete the file when clicked.

There should also be some text in the modal while the connection is being made, currently it's just the "Downloading..." button and then later the progressbar output appears.

guzzisti:
loosely related:
way to often i end up with closing amfora (with all the open tabs) cause `Ctrl+C` is too familiar to stop a action (e.g. loading a page).

I would really prefer to remap `ctrl+c` to stop loading and only `ctrl+q` for closing amfora right away.

makeworld-the-better-one:
I think the majority of people are familiar with <kbd>Ctrl-C</kbd> stopping an application rather than some subroutine like loading a page. In any case, #57 would have to be closed for stopping page loading to be implemented.

guzzisti:
This may be a bit offtopic in this issue, but i'd like to add:
i agree that `ctrl+c` is a often used shortcut to exit a program in a terminal. 
But too often i end up with closing amfora and loose all of my open tabs as i hit `ctrl+c`. It acts too much like `ctrl+q` if theres a difference at all.

comparison to firefox: it uses `ctrl+c` to cancel page loading and `ctrl+q` for exit

makeworld-the-better-one:
Hmm, I didn't know that about Firefox. In Amfora, the main difference is that <kbd>Ctrl-C</kbd> will quit even if you're in the middle of loading or downloading. The other keys will not. This isn't a great behaviour though, it should probably be changed.

In any case, this isn't being worked on right now, when it ends up getting worked on this can be hashed out further.

makew0rld:
I am closing all stale non-bug issues so this issue will be closed. However a PR for this would still be accepted.

PR #134, opened 2020-12-01
Mediatypes support

sudobash1:
This is a partial solution to #121. It uses a slightly different configuration syntax than the one proposed. It looks like this:

``` TOML
[mime-handlers]
# Allows setting the commands to run for various mime types.
# E.g. to open png files with the imagemagic display command:
#   [mime-handlers."image/png"]
#   command = ["display"]
#
# The subtype may be a * to open all files of a set type E.g:
#   [mime-handlers."image/*"]
#   command = ["display"]
#
# Both the type and subtype may be */* to specify the default handler.
#
# If command is "" or [] the file will be opened using the mozz.us http portal
# and a-general.http
#
# Mime type handlers may also have an action field containing one of the following:
# "prompt"   - Ask user every time which action to take.
# "download" - Always download the file without prompting
# "open"     - Always open the file using the below command without prompting
[mime-handlers."*/*"]
command = []
action = "prompt"

[mime-handlers."image/*"]
command = "display"
action = "open"
```

It leaves the mozz.us portal as the default, but it is trivial for a Linux user to add a `*/*` handler for `xdg-open`. On mac I think plain `open` works. I don't know about Windows though.

## Limitations

There are a handful of limitations with this implementation that I want to point out:

* Currently there is no piping/streaming. Files must be downloaded in full before opening
* Files are downloaded to the system temp dir (but this is configurable). They are never removed by amfora. I am relying on the system to clean the tmp folder on reboot.
* MIME parameters are ignored.
* Opening files in a TUI application (like vim/emacs/viu) is a bit tricky. A user would probably need to write a wrapper script

### DIFF
--- a/default-config.toml
+++ b/default-config.toml
@@ -118,6 +118,9 @@ other = 'off'
 #   cmd = ["feh"]
 #   types = ["image/jpeg"]
 #
+# Each command that you specify must come under its own [[mediatype-handlers]]. You may
+# specify as many [[mediatype-handlers]] as you want to setup multiple commands.
+#
 # If the subtype is omitted then the specified command will be used for the
 # entire type:
 #   [[mediatype-handlers]]
@@ -142,9 +145,6 @@ other = 'off'
 #   cmd = ["feh"]
 #   types = ["image"]
 #   no_prompt = true
-#
-# NOTE: There may be multiple [[mediatype-handlers]] specified to setup multiple
-# handler commands
 
 
 [cache]
--- a/default-config.toml
+++ b/default-config.toml
@@ -109,6 +109,13 @@ shift_numbers = "!@#$%^&*()"
 other = 'off'
 
 
+[mime-handlers]
+# Allows setting the commands to run for various mime types.
+# E.g. to open png files with the imagemagic display command:
+#   [mime-handlers."image/png"]
+#   command = "display"
+
+
 [cache]
 # Options for page cache - which is only for text/gemini pages
 # Increase the cache size to speed up browsing at the expense of memory
--- a/default-config.toml
+++ b/default-config.toml
@@ -113,7 +113,7 @@ other = 'off'
 # Allows setting the commands to run for various mime types.
 # E.g. to open png files with the imagemagic display command:
 #   [mime-handlers."image/png"]
-#   command = "display"
+#   command = ["display"]
 
 
 [cache]
--- a/default-config.toml
+++ b/default-config.toml
@@ -109,28 +109,38 @@ shift_numbers = "!@#$%^&*()"
 other = 'off'
 
 
-[mime-handlers]
-# Allows setting the commands to run for various mime types.
-# E.g. to open png files with the imagemagic display command:
-#   [mime-handlers."image/png"]
-#   command = ["display"]
+# [[mediatype-handlers]]
+# Allows setting the commands to run for various mediatypes.
 #
-# The subtype may be a * to open all files of a set type E.g:
-#   [mime-handlers."image/*"]
-#   command = ["display"]
+# To open jpeg files with the feh command:
+#   [[mediatype-handlers]]
+#   cmd = ["feh"]
+#   types = ["image/jpeg"]
 #
-# Both the type and subtype may be */* to specify the default handler.
+# If the subtype is ommited then the specified command will be used for the
+# entire type:
+#   [[mediatype-handlers]]
+#   command = ["vlc", "--flag"]
+#   types = ["audio", "video"]
 #
-# If command is "" or [] the file will be opened using the mozz.us http portal
-# and a-general.http
+# The default catch-all handler can by specified with an "*"
+#   [[mediatype-handlers]]
+#   cmd = ["some-command"]
+#   types = [
+#           "application/pdf",
+#           "*",
+#   ]
 #
-# Mime type handlers may also have an action field containing one of the following:
-# "prompt"   - Ask user every time which action to take.
-# "download" - Always download the file without prompting
-# "open"     - Always open the file using the below command without prompting
-[mime-handlers."*/*"]
-command = []
-action = "prompt"
+# If you want to always open a type in its viewer without the download or open
+# prompt appearing, you can add no_prompt = true
+#
+#   [[mediatype-handlers]]
+#   cmd = ["feh"]
+#   types = ["image"]
+#   no_prompt = true
+#
+# NOTE: There may be multiple [[mediatype-handlers]] specified to setup mutiple
+# handler commands
 
 
 [cache]
--- a/default-config.toml
+++ b/default-config.toml
@@ -123,8 +123,14 @@ other = 'off'
 #
 # If command is "" or [] the file will be opened using the mozz.us http portal
 # and a-general.http
+#
+# Mime type handlers may also have an action field containing one of the following:
+# "prompt"   - Ask user every time which action to take.
+# "download" - Always download the file without prompting
+# "open"     - Always open the file using the below command without prompting
 [mime-handlers."*/*"]
 command = []
+action = "prompt"
 
 
 [cache]
--- a/default-config.toml
+++ b/default-config.toml
@@ -110,20 +110,24 @@ other = 'off'
 
 
 # [[mediatype-handlers]]
-# Allows setting the commands to run for various mediatypes.
+# Specify what applications will open certain filetypes.
+# By default your default application will be used to open the file when you select "Open".
 #
 # To open jpeg files with the feh command:
 #   [[mediatype-handlers]]
 #   cmd = ["feh"]
 #   types = ["image/jpeg"]
 #
-# If the subtype is ommited then the specified command will be used for the
+# If the subtype is omitted then the specified command will be used for the
 # entire type:
 #   [[mediatype-handlers]]
 #   command = ["vlc", "--flag"]
 #   types = ["audio", "video"]
 #
-# The default catch-all handler can by specified with an "*"
+# A catch-all handler can by specified with "*".
+# Note that there are already catch-all handlers in place for all OSes,
+# that open the file using your default application. This is only if you
+# want to override that.
 #   [[mediatype-handlers]]
 #   cmd = ["some-command"]
 #   types = [
@@ -139,7 +143,7 @@ other = 'off'
 #   types = ["image"]
 #   no_prompt = true
 #
-# NOTE: There may be multiple [[mediatype-handlers]] specified to setup mutiple
+# NOTE: There may be multiple [[mediatype-handlers]] specified to setup multiple
 # handler commands
 
 
--- a/default-config.toml
+++ b/default-config.toml
@@ -110,8 +110,10 @@ other = 'off'
 
 
 # [[mediatype-handlers]]
-# Specify what applications will open certain filetypes.
+# Specify what applications will open certain media types.
 # By default your default application will be used to open the file when you select "Open".
+# You only need to configure this section if you want to override your default application,
+# or do special things like streaming.
 #
 # To open jpeg files with the feh command:
 #   [[mediatype-handlers]]
@@ -145,6 +147,14 @@ other = 'off'
 #   cmd = ["feh"]
 #   types = ["image"]
 #   no_prompt = true
+#
+# Note: Multiple handlers cannot be defined for the same full media type, but
+# still there needs to be an order for which handlers are used. The following
+# order applies regardless of the order written in the config:
+#
+# 1. Full media type: "image/jpeg"
+# 2. Just type: "image"
+# 3. Catch-all: "*"
 
 
 [cache]
--- a/default-config.toml
+++ b/default-config.toml
@@ -114,6 +114,17 @@ other = 'off'
 # E.g. to open png files with the imagemagic display command:
 #   [mime-handlers."image/png"]
 #   command = ["display"]
+#
+# The subtype may be a * to open all files of a set type E.g:
+#   [mime-handlers."image/*"]
+#   command = ["display"]
+#
+# Both the type and subtype may be */* to specify the default handler.
+#
+# If command is "" or [] the file will be opened using the mozz.us http portal
+# and a-general.http
+[mime-handlers."*/*"]
+command = []
 
 
 [cache]
--- a/display/download.go
+++ b/display/download.go
@@ -141,7 +141,7 @@ func dlChoice(text, u string, resp *gemini.Response) {
 	if choice == "Open" {
 		tabPages.HidePage("dlChoice")
 		App.Draw()
-		downloadAndOpen(u, resp)
+		open(u, resp)
 		return
 	}
 	tabPages.SwitchToPage(strconv.Itoa(curTab))
@@ -149,10 +149,10 @@ func dlChoice(text, u string, resp *gemini.Response) {
 	App.Draw()
 }
 
-// downloadAndOpen performs the same actions as downloadURL except it also opens the file.
+// open performs the same actions as downloadURL except it also opens the file.
 // If there is no system viewer configured for the particular mediatype, it opens it
 // with the default system viewer.
-func downloadAndOpen(u string, resp *gemini.Response) {
+func open(u string, resp *gemini.Response) {
 	mediaHandler := getMediaHandler(resp)
 	path := downloadURL(config.TempDownloadsDir, u, resp)
 	if path == "" {
--- a/display/download.go
+++ b/display/download.go
@@ -17,7 +17,7 @@ import (
 	"github.com/gdamore/tcell"
 	"github.com/makeworld-the-better-one/amfora/config"
 	"github.com/makeworld-the-better-one/amfora/structs"
-	"github.com/makeworld-the-better-one/amfora/webbrowser"
+	"github.com/makeworld-the-better-one/amfora/sysopen"
 	"github.com/makeworld-the-better-one/go-gemini"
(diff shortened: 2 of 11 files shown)


makeworld-the-better-one:
Thanks for making this PR. This was actually something I wanted to work on personally, which I indicated by assigning myself in #121. Next time feel free to ask and I'll let you know. But I still appreciate the work here, and thanks for your interest in Amfora!

I'll outline some issues I see with the way you've added this below, you're welcome to either make the changes if you agree, or close the PR and I will work on it at some point later.

## Config

- Your new format for configuring mediatypes doesn't seem to allow for specifying multiple mediatypes for one command. A user would have to create multiple sections instead, which seems verbose and repetitive
- I like the idea of being able to disable the prompt modal and just always open the file, but I don't see the value in always initiating a download, especially since there's currently no way to stop one (see #129)
	- Maybe this could be replaced with an optional `no_prompt` variable that defaults to `false`
- `If command is "" or [] the file will be opened using the mozz.us http portal and a-general.http`
	- I don't really see the point of this. Why not just not have the section at all? And the portal usage should be removed entirely, as I mentioned in #121 

> It leaves the mozz.us portal as the default, but it is trivial for a Linux user to add a `*/*` handler for `xdg-open`. On mac I think plain `open` works. I don't know about Windows though.

It's important to me that those commands are the default, built-in to the application. So by default when accessing a file the user gets three options (Open, Download, Cancel), and clicking "Open" will use one of these default handlers. On Windows you can use `exec.Command("rundll32", "url.dll,FileProtocolHandler", file)`, this is used in the `webbrowser/` code as a reference.

Note that opening in the portal was not one of the options, I'd like to remove that dependency.

## Limitations

> Currently there is no piping/streaming. Files must be downloaded in full before opening

This is something I need to have before releasing and before closing #121, but it's fine if it's not in this PR, happy to add it myself.

> Files are downloaded to the system temp dir (but this is configurable). They are never removed by amfora. I am relying on the system to clean the tmp folder on reboot.

That's fine with me, I believe it's what Firefox does as well.

> MIME parameters are ignored.

Also fine.

> Opening files in a TUI application (like vim/emacs/viu) is a bit tricky. A user would probably need to write a wrapper script

I was thinking about this as well. Unfortunately there doesn't seem to be a good way around this, I'm not interested in nesting a terminal inside Amfora as all the implementations of that I've see don't work very well (slow, no truecolor support, etc).

## Code

There are some things in the code I could go over and do a full review, but I'd rather hear your response to this and what you think, as well as see the above changes made, before doing that. 

Thanks again for your work!

sudobash1:
Thanks for the reply. I completely understand wanting to work on this yourself. I initially was just going to kludge something together for myself since I was really wanting to get around using the portal, but I figured it couldn't hurt to clean it a bit and post what I had. Bad form not introducing myself.

Your reply on the config makes sense to me. I'll make changes to this PR. But I'll not be offended in the least if you want to do your own implementation, or rework things.

With regards to the terminal apps, a solution I am using myself is:

```sh
#!/bin/sh
if [ -n "$TMUX" ]; then
  tmux new-window sh -c 'viu "$@"; read -p "[Press Enter]"' sh "$@"
else
  display "$@"
fi
```

Setting this script as the handler will allow you to stay in the terminal if you are using [tmux](https://tmux.github.io/) otherwise it falls back to opening a new window.

Cheers :beers:

makeworld-the-better-one:
>  I initially was just going to kludge something together for myself since I was really wanting to get around using the portal, but I figured it couldn't hurt to clean it a bit and post what I had.

This is how open source is built. Thank you!

> Your reply on the config makes sense to me. I'll make changes to this PR. But I'll not be offended in the least if you want to do your own implementation, or rework things.

Happy to hear it, thanks for understanding. Once you make the changes let me know here, and I'll decide what I want to do. Since you're open to my ideas I suspect I'll probably just end up merging your work.

> With regards to the terminal apps, a solution I am using myself is [snip]

What is that `display` command? Are you on macOS and it's some Apple thing? On my Linux machine it's a command for [displaying images](https://imagemagick.org/script/display.php). Using tmux seems like a cool way to do it though. Thanks for sharing!

sudobash1:
I've pushed an update which uses the configuration you specified. I looked into viper and found its unmarshalling ability made things cleaner.

> What is that display command? Are you on macOS and it's some Apple thing? On my Linux machine it's a command for displaying images. Using tmux seems like a cool way to do it though. Thanks for sharing!

The display command is what you have on Linux (I'm on Linux too). That script is for opening an image. [viu](https://github.com/atanunq/viu) is a program to render an image using utf-8 characters.

> EDIT: Something I forgot mention in my original reply -- please use "mediatype" everywhere instead of "mimetype", it's the newer and more accurate term.

Noted and done. Thanks!

There are a few code smell issues:

#### Location of MediaHandler struct

I wasn't sure where the MediaHandler struct should go. Currently it is `config/config.go` but perhaps it should go in `structs/structs.go`?

#### (Ab)using the webbrowser package

Instead of copying from the `webbrowser` package, I simply used it directly. I think that functionally this is OK, but it seems like an abuse of the webbrowser package in that a) I am not using it for a web browser per se and b) I am discarding the msg it returns, and substituting it for one which doesn't mention "browser". The *nix one is using `$BROWSER` as a fallback, but I don't see that as a particular problem. At least with Firefox or Chrome, if you give it a random file on the command line, it will try to display it, or offer to open/download it.

If it were me, I would be tempted to make the webbrowser package into a general "system open" utility, but I don't want to run off and do my own thing drastically changing stuff (again :wink:). And I may have an overdeveloped aversion to code duplication...

#### Old commits left in place.

The commits should be squashed. I'm leaving things as-is for the PR review.

display/download.go
```diff
@@ -83,56 +86,103 @@ func dlInit() {
...
+	if err != nil {
+		return def
+	}
+
+	if ret, ok := config.MediaHandlers[mediatype]; ok {
+		return ret
+	}
+
+	splitType := strings.Split(mediatype, "/")
+	if ret, ok := config.MediaHandlers[splitType[0]]; ok {
```
makeworld-the-better-one, line 105:
```suggestion
	if ret, ok := config.MediaHandlers[splitType]; ok {
```

display/download.go
```diff
@@ -83,56 +86,103 @@ func dlInit() {
...
 	}
 	tabPages.SwitchToPage(strconv.Itoa(curTab))
 	App.SetFocus(tabs[curTab].view)
 	App.Draw()
 }
 
+// downloadAndOpen performs the same actions as downloadURL except it also opens the file.
+// If there is no system viewer configured for the particular mediatype, it opens it
+// with the default system viewer.
+func downloadAndOpen(u string, resp *gemini.Response) {
```
makeworld-the-better-one, line 155: Can you rename this to just `open` or something? Because in the future this function might not download, but stream/pipe instead.

display/download.go
```diff
@@ -83,56 +86,103 @@ func dlInit() {
...
+		return
+	}
+	if mediaHandler.Cmd == nil {
+		// Open with system default viewer
+		_, err := webbrowser.Open(path)
+		if err != nil {
+			Error("System Viewer Error", err.Error())
+			return
+		}
+		dlModal.SetText("Opened in system default viewer")
```
makeworld-the-better-one, line 168:
```suggestion
		dlModal.SetText("Opened in default system viewer")
```
This sounds more natural to me.

Also I think that this message might be better suited to an `Info`. Do you think you could close the `dlModal` (See lines 80-85 for how to do that) and display this using `Info()` instead? And do the same on line 176.

If that's too messy or you think this method makes more sense let me know.

display/download.go
```diff
@@ -83,56 +86,103 @@ func dlInit() {
...
+	} else {
+		cmd := mediaHandler.Cmd
+		err := exec.Command(cmd[0], append(cmd[1:], path)...).Start()
+		if err != nil {
+			Error("System Viewer Error", "Error executing custom command: "+err.Error())
```
makeworld-the-better-one, line 173:
```suggestion
			Error("File Opening Error", "Error executing custom command: "+err.Error())
```

display/download.go
```diff
@@ -83,56 +86,103 @@ func dlInit() {
...
+			return
+		}
+		dlModal.SetText("Opened in system viewer")
```
makeworld-the-better-one, line 176:
```suggestion
		dlModal.SetText("Opened with "+cmd[0])
```

display/download.go
```diff
@@ -83,56 +86,103 @@ func dlInit() {
...
+	}
+	App.SetFocus(dlModal)
+	App.Draw()
+}
+
 // downloadURL pulls up a modal to show download progress and saves the URL content.
 // downloadPage should be used for Page content.
-func downloadURL(u string, resp *gemini.Response) {
+// Returns location downloaded to or an empty string on error
```
makeworld-the-better-one, line 184:
```suggestion
// Returns location downloaded to or an empty string on error.
```

display/download.go
```diff
@@ -221,13 +273,13 @@ func downloadPage(p *structs.Page) (string, error) {
 // downloadNameFromURL takes a URl and returns a safe download path that will not overwrite any existing file.
 // ext is an extension that will be added if the file has no extension, and for domain only URLs.
 // It should include the dot.
-func downloadNameFromURL(u string, ext string) (string, error) {
+func downloadNameFromURL(dir, u string, ext string) (string, error) {
```
makeworld-the-better-one, line 276:
```suggestion
func downloadNameFromURL(dir, u, ext string) (string, error) {
```

makeworld-the-better-one (changes requested):
Thanks for the changes! Please see my comments below, in this message and the specific code comments.

> Location of MediaHandler struct

I'm fine with it being in `config/config.go`, it seems directly related.

> (Ab)using the webbrowser package

I'm not really okay with this. I think this (mis)usage can result in strange errors, as well as incorrect logic. I'm not in favour of the accidental `$BROWSER` fallback you mention, or that files won't open on *nix systems if there's no Xorg or Wayland display server.

> I would be tempted to make the webbrowser package into a general "system open" utility

I would prefer something like this. But leave the existing webbrowser package in place, as it already has some specific code that I'd rather not shift elsewhere. It should be simple enough to copy `webbrowser/` into a new folder and just remove the stuff that doesn't apply, and change the package name.

> The commits should be squashed. I'm leaving things as-is for the PR review.

Thank you, it's helpful to have to the full history. I will squash the commits for merging.

default-config.toml
```diff
@@ -109,6 +109,40 @@ shift_numbers = "!@#$%^&*()"
...
 
+# [[mediatype-handlers]]
+# Allows setting the commands to run for various mediatypes.
+#
+# To open jpeg files with the feh command:
+#   [[mediatype-handlers]]
+#   cmd = ["feh"]
+#   types = ["image/jpeg"]
+#
+# If the subtype is ommited then the specified command will be used for the
```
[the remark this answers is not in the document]
sudobash1: Can you tell I never did well with spelling at school :upside_down_face:

default-config.toml
```diff
@@ -109,6 +109,40 @@ shift_numbers = "!@#$%^&*()"
...
+# If you want to always open a type in its viewer without the download or open
+# prompt appearing, you can add no_prompt = true
+#
+#   [[mediatype-handlers]]
+#   cmd = ["feh"]
+#   types = ["image"]
+#   no_prompt = true
+#
+# NOTE: There may be multiple [[mediatype-handlers]] specified to setup mutiple
+# handler commands
```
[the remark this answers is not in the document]
sudobash1: Sounds good to me. I agree that it is a bit unintuitive now. It took me an embarrassingly long time to realize that there were two `[[` characters, which meant it was an array. I had to read part way through the TOML documentation before it hit me.

display/download.go
sudobash1: Will do

display/download.go
sudobash1: Aha. I'm glad you have an example of how to do this. I was wondering how to close a modal without user interaction. 

When I was first doing this, I was wondering if the user needed any notification at all. Presumably they will be aware because the download message will have disappeared, and their file has opened up. What do you think?

display/download.go
sudobash1: I thought about doing this, but the reason I didn't was the user may set their command to be something like `xterm -e vim`. If they did this then it would show `Opened with xterm` which seems wrong. Thoughts?

sudobash1:
>> (Ab)using the webbrowser package
>
> I'm not really okay with this. I think this (mis)usage can result in strange errors, as well as incorrect logic. I'm not in favour of the > accidental $BROWSER fallback you mention

Makes sense.

> or that files won't open on *nix systems if there's no Xorg or Wayland display server.

I think that if a user does not have a display server running, then there is not going to be a way to determine the system default viewer. The xdg-open man page says:

> xdg-open is for use inside a desktop session only.

I think the best course of action for when there is no display server is to display an error dialog instructing the user to add their own default `[[mediatype-handlers]]`.

default-config.toml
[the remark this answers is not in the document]
sudobash1: I moved this explanation up below the first example and reworded it slightly.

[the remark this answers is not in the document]
makeworld-the-better-one: Looks good now, thanks.

display/download.go
makeworld-the-better-one: I think that displaying an Info message is good to let the user know Amfora has done its job. Maybe the application that is being called has failed silently, or is taking a long time to open. Displaying an Info shows that any issue that occurs isn't with Amfora.

display/download.go
makeworld-the-better-one: I think showing part of the command is better than not at all. There could be code here that adds all the command parts together with spaces, but I didn't suggest that because I'm worried some commands might be very long and it would be annoying to read. I feel like having a possibly incomplete command is better than that a super long one, or not having one at all.

makeworld-the-better-one:
> I think that if a user does not have a display server running, then there is not going to be a way to determine the system default viewer.

Aha you're totally right, lol. I still think this should be a separate package because of the `$BROWSER` fallback, but otherwise it's basically a copy of `webbrowser/`. I should probably combine them...

sysopen/README.md
```diff
@@ -0,0 +1,5 @@
+# `package sysopen`
+
+The code in this folder is adapted from Bombadillo, you can see the original [here](https://tildegit.org/sloum/bombadillo/src/branch/master/http). Many thanks to Sloum and the rest of the team!
+
+The code is simple, and I have changed it, but in any case there should be no licensing issues because both repos are under GPL v3.
```
makeworld-the-better-one, line 5: I don't think there's any point in having this. If anything, it should say it was adapted from `webbrowser`, which is more accurate. But I think it can just be removed.

sysopen/open_browser_other.go
```diff
@@ -0,0 +1,11 @@
+// +build !linux,!darwin,!windows,!freebsd,!netbsd,!openbsd
+
+package sysopen
+
+import "fmt"
+
+// Open opens `path` in default system viewer, but not on this OS.
+func Open(path string) (string, error) {
+	return "", fmt.Errorf("unsupported OS for default system viewer. " +
+		"Set a default [[mediatype-handlers]] command in the config")
```
makeworld-the-better-one, line 10:
```suggestion
		"Set a catch-all [[mediatype-handlers]] command in the config")
```

sysopen/open_browser_unix.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+	"os/exec"
+)
+
+// Open opens `path` in default system viewer. It tries to do so using
+// xdg-open. It only works if there is a display server working.
+func Open(path string) (string, error) {
+	var (
+		// In prev versions there was also Xorg executable checked for.
+		// I don't see any reason to check for it.
```
makeworld-the-better-one, line 17:
```suggestion
```

sysopen/open_browser_unix.go
```diff
@@ -0,0 +1,37 @@
...
+	case xdgOpenNotFoundErr == nil:
+		// Use start rather than run or output in order
+		// to make application run in background.
+		if err := exec.Command(xdgOpenPath, path).Start(); err != nil {
+			return "", err
+		}
+		return "Opened in default system viewer", nil
+	default:
+		return "", fmt.Errorf("could not determine default system viewer. " +
+			"Set a default [[mediatype-handlers]] command in the config")
```
makeworld-the-better-one, line 35:
```suggestion
			"Set a catch-all [[mediatype-handlers]] command in the config")
```

makeworld-the-better-one (changes requested):
Just some small changes, I think this is almost ready to go! Thanks for this. 

I don't want this in the upcoming v1.7.0 release though, so I might leave this PR open, or merge it into a different branch. It should be in v1.8.0.

default-config.toml
```diff
@@ -109,6 +109,44 @@ shift_numbers = "!@#$%^&*()"
...
+#           "*",
+#   ]
+#
+# If you want to always open a type in its viewer without the download or open
+# prompt appearing, you can add no_prompt = true
+#
+#   [[mediatype-handlers]]
+#   cmd = ["feh"]
+#   types = ["image"]
+#   no_prompt = true
```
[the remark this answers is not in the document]
sudobash1: Good idea to document this.

sudobash1:
OK. The CI checks have passed, and I think I got all of your requests in. :beers:

makeworld-the-better-one:
Alright this looks done, thank you! I'll merge it into the new `mediatype-handlers` branch, and then it should be in the v1.8.0 release.

PR #135, opened 2020-12-04
Keybinding system for global keys

phaedrus-jaf:
I've been really enjoying using amfora, but it has different keybindings from qutebrowser and my terminal translates ctrl-t to ctrl-a,c to get screen to create a new tab/view.  Also, my compositor eats all F1 keypresses.  Re-doing the hardcoding has been nice for a couple weeks, but that doesn't let me keep up to date with y'all.

I think this code is pretty straightforward, let me know what you think or if you wanted a totally different design.  (I know I didn't ask, I just wrote something that worked for me :)


makeworld-the-better-one:
Hello, happy to hear you've been enjoying Amfora! Thanks for making this PR, keybindings is a good feature, and it's even better when it's not just on your fork :)

I think your code is good, but I find the format you chose kind of ugly/unclear to write. Using `Rune:` is verbose, it's weird to have `Alt:` but `Ctrl+`, etc. I was thinking you could take a look at cbind's [key.go](https://gitlab.com/tslocum/cbind/-/blob/master/key.go) file, for an idea of how decoding would work. I find the way cbind encodes keybindings much cleaner and more intuitive. Feel free to copy that code entirely, it's under the MIT license. Note I'm not looking for the entire functionality of cview to be copied, just the ideas in that one file.

Some other things:
- Use TOML arrays instead separating keybindings with commas. This allows for easily binding commands to the comma key, as well as just simpler parsing
- Please keep `shiftnumbers` in there. It's ugly, but I don't want to break existing configs.
- Make sure you're not exporting functions/variables/structs that aren't used elsewhere

phaedrus-jaf:
Good point about the TOML parser...  I feel dumb for not seeing that.

I decided to go with `Alt-` and keep `Ctrl-`, because it required only minor tweaks to what I had written (and it's simpler than the cbind code).  I got rid of the 'Rune:' business.  I also decided to handle `Space` as a special case.  If we want to be able to have whitespace in there, a couple `strings.ReplaceAll` calls would allow that.

And finally, I re-added `shift_numbers`, because not breaking non-US configs is a good idea.  The code I wrote does enforce 10 characters, where the old code did not.  I might go back and make it more liberal in a bit here, I want to clear the merge conflict first.

phaedrus-jaf:
Ok, I think this should be ready to merge.

phaedrus-jaf:
Oh!  I just realized something that should probably go with this: the help screen is static.  I'll take a shot at updating it to display the configured keys later tonight. (I'm also happy to update the wiki to add a keybindings page under the configuration page)

makeworld-the-better-one:
Nice catch, thanks. I'll take a look at everything when I can, after you update the help page.

makeworld-the-better-one:
Looks like maybe that part is done? Let me know and I'll take a look soon.

phaedrus-jaf:
Yes, I got the help page working, and I think it's ready.

config/config.go
```diff
@@ -247,6 +286,9 @@ func Init() error {
 		return err
 	}
 
+	// Setup the key bindings:
```
makeworld-the-better-one, line 289:
```suggestion
	// Setup the key bindings
```

default-config.toml
```diff
@@ -87,12 +87,60 @@ emoji_favicons = false
...
+#bind_tab8 = "*"
+#bind_tab9 = "("
+#bind_tab0 = ")"
+# Also, shift_numbers still works:
 # Hold down shift and press the numbers on your keyboard (1,2,3,4,5,6,7,8,9,0) to set this up.
-# It is default set to be accurate for US keyboards.
-shift_numbers = "!@#$%^&*()"
-
+# EG, for US keyboards (the default):
+#shift_numbers = "!@#$%^&*()"
```
makeworld-the-better-one, line 105: `shift_numbers` no longer needs to be documented, since it's deprecated.

default-config.toml
```diff
@@ -87,12 +87,60 @@ emoji_favicons = false
...
+
+# Notes:
+# * whitespace is not allowed in any of the keybindings! Use 'Space' and 'Tab' to bind to those keys.
+# * Multiple keys can be bound to one command, just use a TOML array.
+# * To add the Alt modifier, the binding must start with Alt-, should be reasonably universal
+# * Ctrl- isn't a universal modifier, just the keys listed in tcell.KeyNames, see:
+# https://github.com/gdamore/tcell/blob/master/key.go (it was line 83 when this comment was written)
```
makeworld-the-better-one, line 112:
```suggestion
# Whitespace is not allowed in any of the keybindings! Use 'Space' and 'Tab' to bind to those keys.
# Multiple keys can be bound to one command, just use a TOML array.
# To add the Alt modifier, the binding must start with Alt-, should be reasonably universal
# Ctrl- won't work on all keys, see this for a list:
# https://github.com/gdamore/tcell/blob/cb1e5d6fa606/key.go#L83
```

default-config.toml
```diff
@@ -87,12 +87,60 @@ emoji_favicons = false
...
+
+# And the rest of the default keybindings
```
makeworld-the-better-one, line 114: I don't think it's necessary to list all the default keybindings here. See how the `[theme]` section is done, where all the keys are listed, and some are explained, but not all the values. So maybe explain `bind_edit`, the `bind_link*` ones, but for the rest just list them.

config/keybindings.go
```diff
@@ -1,24 +1,216 @@
...
+	CmdTab2    = 12
+	CmdTab3    = 13
+	CmdTab4    = 14
+	CmdTab5    = 15
+	CmdTab6    = 16
+	CmdTab7    = 17
+	CmdTab8    = 18
+	CmdTab9    = 19
+	CmdTab0    = 20
+	CmdBottom  = iota
```
makeworld-the-better-one, line 32: Is there any reason why you did it like this? You can just specify `= iota` at the start.

config/keybindings.go
```diff
@@ -1,24 +1,216 @@
...
-// The error is not nil if the provided key is invalid.
-func KeyToNum(key rune) (int, error) {
-	runes := []rune(viper.GetString("keybindings.shift_numbers"))
-	for i := range runes {
-		if key == runes[i] {
-			if i == len(runes)-1 {
-				// Last key is 0, not 10
-				return 0, nil
+const (
+	CmdInvalid = 0
```
makeworld-the-better-one, line 11: I think the code would be more clear if you defined something like `type Command int`, and made all these consts of that type. Then later lines like `func GetKeyBinding(cmd int) string` would become `func GetKeyBinding(cmd Command) string` which I find more readable and foolproof.

config/keybindings.go
```diff
@@ -1,24 +1,216 @@
...
+)
+
+type keyBinding struct {
+	key tcell.Key
+	mod tcell.ModMask
+	r   rune
+}
+
+var bindings map[keyBinding]int
+var tcellKeys map[string]tcell.Key
```
makeworld-the-better-one, line 60: Please add some comments explaining each of these are.

config/keybindings.go
```diff
@@ -1,24 +1,216 @@
...
+
+func keyBindingToString(kb keyBinding) (string, bool) {
```
makeworld-the-better-one, line 62: Please document this function and all the ones below it.

config/keybindings.go
```diff
@@ -1,24 +1,216 @@
...
-	return -1, errors.New("provided key is invalid") //nolint:goerr113
+}
+
+func TranslateKeyEvent(e *tcell.EventKey) int {
+	var ok bool
+	var cmd int
+	k := e.Key()
+	if k == tcell.KeyRune {
+		cmd, ok = bindings[keyBinding{k, e.Modifiers(), e.Rune()}]
+	} else { // Sometimes tcell sets e.Rune() on non-KeyRune events.
```
makeworld-the-better-one, line 209: Where have you seen this? Just curious.

display/display.go
```diff
@@ -245,43 +245,36 @@ func Init() {
 			return event
 		}
 
+		/* To add a configurable global key command, you'll need to update one of
+		   the two switch statements here.  You'll also need to add an enum entry in
+		   config/keybindings.go, update KeyInit() in config/keybindings.go, add a default
+		   keybinding in config/config.go and update the help panel in display/help.go
+		*/
```
makeworld-the-better-one, line 252:
```suggestion
		// To add a configurable global key command, you'll need to update one of
		// the two switch statements here.  You'll also need to add an enum entry in
		// config/keybindings.go, update KeyInit() in config/keybindings.go, add a default
		// keybinding in config/config.go and update the help panel in display/help.go
		
```
Go code tends to avoid the block style comments.

display/help.go
```diff
@@ -71,6 +71,35 @@ func helpInit() {
...
+		config.GetKeyBinding(config.CmdHome),
+		config.GetKeyBinding(config.CmdNewTab),
+		config.GetKeyBinding(config.CmdCloseTab),
+		config.GetKeyBinding(config.CmdReload),
+		config.GetKeyBinding(config.CmdBookmarks),
+		config.GetKeyBinding(config.CmdAddBookmark),
+		config.GetKeyBinding(config.CmdSave),
+		config.GetKeyBinding(config.CmdSub),
+		config.GetKeyBinding(config.CmdAddSub),
+		config.GetKeyBinding(config.CmdQuit))
```
makeworld-the-better-one, line 101:
```suggestion
		config.GetKeyBinding(config.CmdQuit),
		)
```

makeworld-the-better-one (changes requested):
Thank you for working on this! Overall it looks good, I've made some specific comments. Also:

- In general, put a space after the hashtag in the config file. So `#bind_whatever` becomes `# bind_whatever`.
- Add comments to document functions

Also, are you interested in writing a [wiki](https://github.com/makeworld-the-better-one/amfora/wiki) page on this? It would be under the configuration section, like the [Proxying page](https://github.com/makeworld-the-better-one/amfora/wiki/Proxying).

config/keybindings.go
phaedrus-jaf: It was meant as a semantic warning: there's some code that depends on these values being in this relation to each other.  Essentially: CmdTab0 - CmdTab1 == 9, and CmdTab[N+1] == 1 + CmdTabN, thus rearranging these would break stuff.  Basically, get future programmers to ask, wait, why does this break convention?

I'll add a comment to this effect.

config/keybindings.go
phaedrus-jaf: Primarily on the Ctrl-Letter-Key events.  The event will have `e.Key() == KeyCtrlA`, `e.Modifers() == 2` (for Control) and `e.Rune() == 'A'`.  I originally added a correction for this when generating the bindings table, but it felt awkward, and would break if tcell changed behavior.

phaedrus-jaf:
Just commited the suggestions that could be committed in the GitHub UI.  I'll go work through the rest of the conversations in a bit here.  (And fix the linter error)

phaedrus-jaf:
And I think I mentioned it, but I'm definitely happy to update the wiki and add a page for keybindings under the configuration page.

phaedrus-jaf:
I think I addressed all the comments and the linter is happy.  It probably makes sense to hold off on trying to update the wiki until after the merge.

config/keybindings.go
makeworld-the-better-one: Alright, that makes sense. And yeah a comment is helpful, thanks.

makeworld-the-better-one:
Thanks for adding this! I'm having a little trouble grokking it but I think I'm just a bit tired right now. Overall it looks good.

PR #150, opened 2020-12-21
Update Nord theme for 1.7.0

lokesh-krishna:
Adds colors for the new subscription modal


makeworld-the-better-one:
Thanks for updating! While I still see the image, likely because of GitHub's caching, the link you provided, `https://github.com/lokesh-krishna/images/blob/main/Swappshot%20Mon%20Dec%2021%2014:37:18%202020.png?raw=true`, doesn't work.

lokesh-krishna:
My bad. I had my repository set to private and I think that might have been the cause of the issue? I updated the link now.

makeworld-the-better-one:
Looks good, thanks.

PR #152, opened 2020-12-21
📝 add kiss linux instructions

jedahan:


makeworld-the-better-one:
Thanks for adding this!

PR #180, opened 2021-02-05
Made tab the delimiter for the help message table.

### LINKED ISSUE #179: Help screen breaks if you bind | character
I took a glance in `help.go` and couldn't help but wonder what would happen if I bound the `|` character to something. Much breakage as it turns out.

With this config:

``` toml
[keybindings]
bind_home = "|"
```

The help screen looks like this:

image

I will see if I can find a simple fix for this.

sudobash1:
This fixes the display issue when the `|` character is used as a keybinding #179 

I used literal tab characters in the help template. Since it is a raw string `\t` would not have worked.

Fixes #179

### DIFF
--- a/display/help.go
+++ b/display/help.go
@@ -10,44 +10,43 @@ import (
 	"gitlab.com/tslocum/cview"
 )
 
-var helpCells = strings.TrimSpace(`
-?|Bring up this help. You can scroll!
-Esc|Leave the help
-Arrow keys, h/j/k/l|Scroll and move a page.
-%s|Go up a page in document
-%s|Go down a page in document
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
Thanks! But wait, does it need to be a raw string? Could you try and put it back as a regular string, and use `\t`? That would help prevent future additions from being added incorrectly with spaces instead of tabs.

sudobash1:
It works fine. It's not as pretty but is perhaps more understandable. Changed and pushed.

makeworld-the-better-one:
Thanks!

Issue #149, opened 2020-12-21
Amfora themeing updates


makeworld-the-better-one:
This thread exists so that everyone who has created a theme for Amfora can know when they need to update it. When Amfora adds a new themeable object, like a new modal, it will be announced in this thread. Users who have created themes can subscribe to this thread, and update their themes accordingly.

makeworld-the-better-one:
@sergetymo @bnthor @Skraylet @crdpa @lokesh-krishna You have all contributed themes to Amfora. Please see the above message and subscribe to this thread if you'd like to keep your theme maintained as Amfora improves. Thanks!

makeworld-the-better-one:
For example, v1.7.0 is about to be released. It changes `preformatted_text` to a light yellow by default, so if you didn't explicitly set that value it will now look different. It also adds the `subscription_modal_bg` and `subscription_modal_text` values.

I am wary of changing your themes myself, so if you're interested it'd be appreciated if you made PRs to update them. Thanks! And no worries if you aren't interested in maintaining. Feel free to let me know if you rather I made the updates instead.

crdpa:
Unfortunately i'm not using that theme anymore (Dracula) so if you can update it would be fine for me. The preformatted_text was already set in the original theme.

makeworld-the-better-one:
@crdpa Okay, thanks for letting me know. I've updated the theme for the subscription modal using those values. This can be seen in d710085. I'll try and maintain the Dracula theme from now on, and you're welcome to take it back at any time by letting me know here.

crdpa:
Thanks. I'll probably come back soon.

Throwing an idea here. Wouldn't be better if amfora could do something like #include in it's config file?

Users could alternate themes just changing one line at the config.

sergetymo:
Made required additions and changes in #153.

makeworld-the-better-one:
@crdpa Hmm, an include key could be a good idea. Maybe like this:

```toml
[theme]
include = "path/to/file.toml"

# Other keys will override the theme
```

Or were you thinking the key could point to a network location?

PR #204, opened 2021-02-25
Add greyscale theme

leifmetcalf:


makeworld-the-better-one:
Thanks for adding this! I never thought about a grayscale theme, but it's a cool idea.

Please subscribe to #149 so you can be updated if any changes to the theme are required.

PR #209, opened 2021-03-07
ability to specify keys to highlight next/prev link

howeyc:
any interest in adding someting like this?

I know there's likely more change required and perhaps this is not exactly the way it should be done, but perhaps it's a starting point?

I'm not sure how to replace the current dependency on the Tab key, or if that's desired.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -220,6 +220,8 @@ func Init() error {
 	viper.SetDefault("keybindings.bind_close_tab", "Ctrl-W")
 	viper.SetDefault("keybindings.bind_next_tab", "F2")
 	viper.SetDefault("keybindings.bind_prev_tab", "F1")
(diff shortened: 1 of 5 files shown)


makeworld-the-better-one (commented):
The problem is that [`TextView.SetDoneFunc`](https://docs.rocketnine.space/gitlab.com/tslocum/cview/#TextView.SetDoneFunc) (see line 49 of `display/tab.go`):

> is called when the user presses on the following keys: Escape, Enter, Tab, Backtab

So that would have to be changed upstream with cview before this could happen. It would probably be a good thing though.

howeyc:
Oh, I see. So removing Tab, or assigning Tab to another command is not currently possible.

However, the change I made could allow for another configurable key to be assigned to next/prev link highlighting.

howeyc:
True. Fair enough.

Issue #212, opened 2021-03-21
Allow custom URL placement for external handlers


t-900-a:
Relevant code: https://github.com/makeworld-the-better-one/amfora/blob/4ecdede58006e77dd23a27b40752103afa6fb989/display/handlers.go#L81

It would be better if the custom commands that are specified in the config could have parameters within them.

Right now the url is concatenated to the end of the specified command.

I.e. instead of appending to the end, all instances of {url_param} are replaced within the url

makeworld-the-better-one:
Could you give me an example use case where having the URL at the end of the command doesn't work?

t-900-a:
> Could you give me an example use case where having the URL at the end of the command doesn't work?

Opening up a new terminal window to display something in a specific application
`
gnome-terminal -- "myprogram --uri %s"`


I don't see a way to display a cli program in a new terminal without having the quotes

I'd definitely be open to any alternatives, if there aren't any would you be willing to accept a pull request with this change?

makeworld-the-better-one:
There might be a way around this, but if you're offering a pull request than I'm happy to merge this and add it. Just make sure you document it in the `default-config.toml` file. I would suggest something like `{{url}}` being the replaced text.

makew0rld:
I am closing all stale non-bug issues so this issue will be closed. However a PR for this could still be accepted.

PR #213, opened 2021-03-28
Allow custom URL placement for external handlers

t-900-a:
Previously the URl was appended to the end now it can be placed anywhere using {{url}}
https://github.com/makeworld-the-better-one/amfora/issues/212

### DIFF
--- a/display/handlers.go
+++ b/display/handlers.go
@@ -85,7 +85,7 @@ func handleOther(u string) {
 			Error("URL Error", "URL contains semicolon, which could indicate malicious intent")
 		} else {
 			for i, field := range fields[1:] {
-				fields[i+1] = strings.Replace(field, "{{url}}", u, -1)
+				fields[i+1] = strings.ReplaceAll(field, "{{url}}", u, -1)
 			}
 			err := exec.Command(fields[0], fields[1:]...).Start()
 			if err != nil {
--- a/display/handlers.go
+++ b/display/handlers.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"os/exec"
 	"path"
+	"regexp"
 	"strings"
 
 	"github.com/makeworld-the-better-one/amfora/cache"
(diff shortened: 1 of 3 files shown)


display/handlers.go
```diff
@@ -78,9 +79,18 @@ func handleOther(u string) {
 	default:
 		// The config has a custom command to execute for URLs
 		fields := strings.Fields(handler)
-		err := exec.Command(fields[0], append(fields[1:], u)...).Start()
-		if err != nil {
-			Error("URL Error", "Error executing custom command: "+err.Error())
+		// safety check for semicolon to prevent url from escaping out of custom command
```
makeworld-the-better-one, line 82: There is no need for a safety check in my opinion. In most cases it will not be necessary, because the URL and other arguments are not being sent to the shell, they are being sent to a binary to use. I understand how in [your use case](https://github.com/makeworld-the-better-one/amfora/issues/212#issuecomment-804360604) a semicolon could mess things up, but I think that's a rare case, and additionally there are going to be many ways to try and escape a command, and it would be tedious, complex. and error-prone to try and enumerate all of them here. We don't even know what shell the user is using.

For your specific case, you should quote the URL: `gnome-terminal -- "myprogram --uri '{{url}}'"`

makeworld-the-better-one (changes requested):
Thanks for working on this! The main problem is that you've broken compatibility with older configs. Specifying a command without `{{url}}` should still work, and use the old behaviour.

t-900-a:
> Thanks for working on this! The main problem is that you've broken compatibility with older configs. Specifying a command without `{{url}}` should still work, and use the old behaviour.

Ok, I'll do a regex match for {{url}} if it's not found, then I'll append to the end as was done previously.

We'll look at this on the weekend

display/handlers.go
t-900-a: I'm probably going to url encode specific special characters. i.e. ' " ; 
There has to be a safety check here.

Someone could easily wget an executable, then run the executable

makeworld-the-better-one: Hmm, this is getting tricky. But all you should need is to escape `'`. And then in the wiki and in `default-config.toml` we'll recommend using single quotes. The only way to do something malicious from within single quotes in bash is to close the quote and do something else, so you only need ot escape `'`. Right?

I feel like a simpler way of doing this needs to be found... One that isn't specific to shells.

t-900-a: > I feel like a simpler way of doing this needs to be found... One that isn't specific to shells.

Found: https://linux.die.net/man/1/xdg-open
I spent some time trying to pwn myself this way. I haven't found a way.

t-900-a:
> Thanks for working on this! The main problem is that you've broken compatibility with older configs. Specifying a command without `{{url}}` should still work, and use the old behaviour.

fixed

makeworld-the-better-one:
Thanks for updating! Two things:

- You can just use `strings.Contains` instead of a regex
- If you want to also close #214 that could be combined in this PR. Up to you.

makeworld-the-better-one:
@t-900-a any update on this?

makew0rld:
Closing manually due to inactivity and lack of Amfora development.

Issue #197, opened 2021-02-17
Remove left margin when scrolling


makeworld-the-better-one:
Due to the change in 90654cd2cfdc4e6cf7bb2cdd35b1171c4d7807e9 (as a part of #107, to fix #26) the left margin on pages was made into a UI element. That means that when scrolling to the right, for long pre-formatted lines, they appear to "disappear" under the left margin. This is an undesirable effect.

One solution would be to change the size of the margin when scrolling, and eventually remove it once its width becomes <= 0. Another possible solution would be to find a way to keep the left margin "underneath" the text.

makeworld-the-better-one:
cc @objectliteral for originally bringing this up

makeworld-the-better-one:
This can probably be done by using `SetInputCapture` on the `TextView`, and changing the margin size if scroll keys are being used. That would be left and right arrow, as well as h and l.

makeworld-the-better-one:
This is almost done, and the work is on the `issue-197` branch. cview will need to be modified to continue. I will make the modifications and switch to using my fork, and then make a PR for it.

makeworld-the-better-one:
Branch has been merged, this issue is fixed! I opened [this](https://gitlab.com/tslocum/cview/-/merge_requests/14) upstream PR.

makeworld-the-better-one:
This is still broken. If you go to `gemini://makeworld.space/test/wrapping-test.gmi` and scroll to the bottom, then start scrolling to the right, it will jump back up to the longest line.

makeworld-the-better-one:
Also: will `pageUp` and `pageDown` affect this?

makeworld-the-better-one:
Alright so scrolling to the bottom with arrow keys was fine, because that's captured. It's scrolling with PgUp or PgDown, or non-capture keys like g and G that was the issue. Now all those keys are captured, and so the internal scroll position is maintained no matter what key is used.

This was fixed in ed36eee.

makeworld-the-better-one:
`ScrollToHighlight` is called for link navigation, and will be called for search (#240). This can mess up the custom scroll logic a lot. A workaround is needed.

makeworld-the-better-one:
Fixed by 6ec7ed1.

PR #222, opened 2021-04-11
Add option for custom keybinds for navigation

### LINKED ISSUE #211: Set keybindings for up/down... movement
I don't notice a way to set my own bindings for scroll up, down (and probably some other basic movements such as left, right too).

Sorry if I'm missing a way to set these, but it seems they must be hardcoded somewhere to arrow keys and hjkl. 

Optionally setting these keys would be helpful for users with alternative keyboard layouts, or users may just have some other preference for how to move about.

singalhimanshu:
Fixes #211


display/help.go
```diff
@@ -78,6 +78,10 @@ func helpInit() {
 		strings.Split(config.GetKeyBinding(config.CmdLink0), ",")[0])
 
 	helpCells = fmt.Sprintf(helpCells,
+		config.GetKeyBinding(config.CmdMoveup),
+		config.GetKeyBinding(config.CmdMovedn),
+		config.GetKeyBinding(config.CmdMoveLeft),
+		config.GetKeyBinding(config.CmdMoveRight),
```
makeworld-the-better-one, line 84: You need to add a place for these arguments to go in the `helpCells` text.

display/tab.go
```diff
@@ -128,7 +128,8 @@ func makeNewTab() *tab {
 
 		key := event.Key()
 		mod := event.Modifiers()
-		ru := event.Rune()
+		// ru := event.Rune()
+		cmd := config.TranslateKeyEvent(event)
```
makeworld-the-better-one, line 132: Since the whole `event` is being looked at, get rid of the `key` and `mod` variables, as well as the `ru` comment. Then fix the logic below to just look at `cmd`.

display/tab.go
```diff
@@ -158,26 +159,30 @@ func makeNewTab() *tab {
 				}
 			}
 			t.page.Column++
+			t.view.ScrollTo(t.page.Row, t.page.Column)
```
makeworld-the-better-one, line 162: None of these added lines should be needed, please test without them. The actual scroll of the `TextView` is changed in `applyHorizontalScroll` and `applyScroll`.

makeworld-the-better-one (changes requested):
Thanks for creating this PR! Please see my comments.

- Add defaults to `config/config.go`
- Test with pages that fit entirely on the screen, as well as pages that are too long, too wide, and both. Make sure scrolling works as expected. Keep in mind that the current scrolling code in master (not in v1.8.0) is slightly broken, see [this comment](https://github.com/makeworld-the-better-one/amfora/issues/197#issuecomment-812582553). If this code reveals a new error let me know, and we can figure out if it's this PR or in master.

makeworld-the-better-one:
Note to self: update wiki keybindings once this is merged

config/default.go
```diff
@@ -148,6 +148,10 @@ scrollbar = "auto"
 # bind_reload
 # bind_back
 # bind_forward
+# bind_moveup
+# bind_movedn
+# bind_moveleft
+# bind_moveright
```
makeworld-the-better-one, line 154: Remove these from this file. Add them to `default-config.toml` instead, and then run `go generate ./...`. Sorry this is not documented.

display/tab.go
singalhimanshu: why aren't we using `applyScroll` instead of `applyHorizontalScroll` here https://github.com/makeworld-the-better-one/amfora/blob/d4c7e87838d2ad2cc58d3b56995ee456a614e238/display/tab.go#L188

makeworld-the-better-one: See the comments at the beginning of the section: https://github.com/makeworld-the-better-one/amfora/blob/d4c7e87838d2ad2cc58d3b56995ee456a614e238/display/tab.go#L126-L127

Whenever vertical scrolling happens, the keys are returned and the actual scrolling is handled by cview. If that needs to be changed it can, but it shouldn't matter for this PR.

singalhimanshu: But this will cause issues when the user will change the `moveup` and `movedown` key-binds to something other than `k` and `j`. In that case cview won't handle the scrolling, it should be done manually by calling `applyScroll` and increment/decrement the value of `t.Page.Row`
One more issue is that `t.Page.Row` starts from 0 which means that even if we reach at the end of the page the value of `t.page.Row` can still increment.
This can be resolved by using `termH` and doing some mathematical operations to make sure that we don't overflow the value of `t.page.Row` beyond the `height`.

config/default.go
singalhimanshu: I have added that to the `default-config.toml` `go generate ./...` generates this default.go so I guess no changes are required here.

display/tab.go
makeworld-the-better-one: Yes, you're correct. Those changes all sound good to me. That should be implemented in this PR. It means the `return`s will have to changed to `return nil` instead of `return event`, so the `TextView` (`t.view`) doesn't see the up and down keys.

config/default.go
makeworld-the-better-one: Yep, seems good then.

makeworld-the-better-one:
See my comments and let me know once you think all the required changes have been made. Thanks!

display/tab.go
singalhimanshu: @makeworld-the-better-one There's one more problem, the help page doesn't implement any custom scrolling it depends upon the scrolling from`cview`. Should I implement that as well in this PR?

singalhimanshu: @makeworld-the-better-one any update on what should I do about the help page?

makeworld-the-better-one: > But this will cause issues when the user will change the `moveup` and `movedown` key-binds to something other than `k` and `j`. In that case cview won't handle the scrolling, it should be done manually

Ok I actually figured out a solution to this, and it applies to help page too. cview has defaults for scrolling keys, but these can be overridden. See [here](https://docs.rocketnine.space/code.rocketnine.space/tslocum/cview#pkg-variables) and scroll down to `var Keys =`.

So basically what should be done is handle left/right scrolling keys here. For up/down scrolling, set the keys in the `cview.Keys` map before starting, like in the `config/` package. You should also set the left/right scroll keys, because despite them being captured here, they're needed for the help page as you pointed out.

Does that make sense? Thanks for waiting.

singalhimanshu: Seems like a good solution. I have implemented this you can review that and let me know if they are in correct place or not.

makeworld-the-better-one: Looks good!

makeworld-the-better-one:
Thanks for updating this! Merged. I made some changes of my own at the end there. The only important one is dd83115. I added back the up/down scrolling lines and changed some of it to use the commands. I also track arrow keys for all scrolling in this section. That's to match the scrolling behaviour in the master branch, which requires tracking the scrolling in all directions, no matter what key is being used. You can see #197 for more details about this scroll behaviour.

Issue #220, opened 2021-04-10
Yank current page URI and/or currently highlighted URI


lostleonardo:
The last time I proposed a UI change, it turned out that Amfora already did it. :-)

With that in mind, is it possible to yank the current URI and put it on the clipboard, without first going into edit mode and mousing over the selection (I would suggest 'yy')? Or, similarly, yank currently highlighted URI ('y')? 

If you like the idea and can point me in the right direction, I could probably have a look at implementing this.

makeworld-the-better-one:
This is not implemented right now. Happy to hear you're interested in adding it yourself.

Adding a new keybinding requires editing `default-config.toml`, `config/keybindings.go`, `config/config.go`, and `display/help.go`.

For yanking the current URI:

- Adding a shortcut that requires timing, like "yy", will be hard if not impossible with the current setup. You should pick a different one.
- To actually activate the keybinding, you'll want to add a case to [this switch](https://github.com/makeworld-the-better-one/amfora/blob/67126ae7855a4b70e723c5f868931fe6a9bfb19d/display/display.go#L297)
- Then in the `case` you can get the `bottomBar` text and put it into the clipboard with a cross-platform clipboard library. Ideally one that doesn't require CGo.

As for yanking the currently highlighted URI (good idea):

- Create a func in `display/tab.go` that returns the currently selected URI, see [these lines](https://github.com/makeworld-the-better-one/amfora/blob/67126ae7855a4b70e723c5f868931fe6a9bfb19d/display/tab.go#L77-L85) for how
- Then add a case to the switch mentioned above that calls that func and uses the clipboard library

makeworld-the-better-one:
Added in #225

PR #225, opened 2021-04-12
Adding bindings to yank the current URL and the selected URL 

lostleonardo:
Many thanks for the excellent support regarding where to make the changes - and for developing a code base that allowed me to get something working with so little difficulty. :-)

I used the following clipboard library https://pkg.go.dev/github.com/atotto/clipboard?utm_source=godoc ... I don't think it needs cgo, although it does depend on xsel or xclip (and their equivalents on Windows and Mac OS). Hopefully that fits with what you were thinking.

Issue #220


singalhimanshu:
I guess you should add `bind_yank_page_uri"`, and `bind_yank_target_uri"` under https://github.com/makeworld-the-better-one/amfora/blob/fe35f45e8c337c8feb6fb21fd0abf7a744995dc8/config/default.go#L160
Otherwise `~/.config/amfora/config.toml` won't have these options in default config.

lostleonardo:
Thanks @singalhimanshu. Now added.

display/tab.go
```diff
@@ -323,3 +323,15 @@ func (t *tab) applyAll() {
 		t.applyBottomBar()
 	}
 }
+
+// GetCurrentURI returns the currently selected URI
+func GetCurrentURI() string {
```
makeworld-the-better-one, line 328: Instead of making this a global function, make it a method of `tab` called `HighlightedURL`, and then call it like `tabs[curTab].HighlightedURL()`.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	code.rocketnine.space/tslocum/cview v1.5.4
+	github.com/atotto/clipboard v0.1.4 // indirect
```
makeworld-the-better-one, line 7: Make sure to run `go mod tidy` to get dependencies in order. Running that should remove the `// indirect` part of this line.

makeworld-the-better-one (changes requested):
Thanks for creating this PR! See my comments below.

Sorry I didn't mention this in #220, but I using URL is better than URI, and that using the word copy is better than yank. I would recommend these two names in the config:

```
bind_copy_page_url
bind_copy_highlight_url
```

Get rid of the `addToClipboard` function, and instead call 	`clipboard.WriteAll` directly. Capture the error, and

go.mod
lostleonardo: I did wonder about the naming. Now updated to match the existing naming convention. I think I caught all of them.

Hopefully, I am now capturing the error in `display/display.go` in an appropriate way.

makeworld-the-better-one: This part looks good. I will review the rest and let you know.

display/display.go
```diff
@@ -352,6 +353,42 @@ func Init(version, commit, builtBy string) {
...
+					err := clipboard.WriteAll(selectedURL)
+					if err != nil {
+						Error("Copy Error", err.Error())
+						return nil
+					}
+					return nil
+				} else if strings.HasPrefix(selectedURL, "/") {
+					// Handle relative URLs
+					u, _ := url.Parse(currentURL)
+					err := clipboard.WriteAll(u.Scheme + "://" + u.Host + selectedURL)
```
makeworld-the-better-one, line 378: Good thinking handling absolute and relative URLs. However there are problems with this approach. For example, if the selected URL is `mailto:test@example.com`. Or if the selected URL is relative but doesn't start with a slash, like `files/test.png`.

Instead of this, you should parse the `currentURL`, and then use [`.Parse`](https://golang.org/pkg/net/url/#URL.Parse) to get the absolute URL. If there's an error, just copy the text of the `selectedURL`.

display/display.go
```diff
@@ -352,6 +353,42 @@ func Init(version, commit, builtBy string) {
...
+				currentURL := tabs[curTab].page.URL
+				err := clipboard.WriteAll(currentURL)
+				if err != nil {
+					Error("Copy Error", err.Error())
+					return nil
+				}
+				return nil
+			case config.CmdCopyTargetURL:
+				currentURL := tabs[curTab].page.URL
+				selectedURL := tabs[curTab].HighlightedURL()
```
makeworld-the-better-one, line 366: You should handle the case where `selectedURL` is an empty string, indicating nothing is selected. I would just do nothing in that case, `return nil`.

makeworld-the-better-one (changes requested):
Thanks for updating! See my comments below.

display/tab.go
lostleonardo: Updated.

display/display.go
lostleonardo: Good call. Updated to make better use of the Go standard library.

makeworld-the-better-one:
Looks great! I tested and it all works fine. Thanks!

Issue #231, opened 2021-04-22
Open files with relative path


makeworld-the-better-one:
Right now, Amfora can open files with a command like `amfora file:///absolute/path/to/file`. But it'd be nice if like the `firefox` command, it could just open files and folders with a relative path, like `amfora file.txt`.

The code would first check if the file or folder exists, and then if not interpret it as a URL.

singalhimanshu:
I can work on this if no one else is already on it.

makeworld-the-better-one:
Sounds good! I'll assign it to you. If you end up being unable to work on it for whatever reason, just let me know.

PR #232, opened 2021-05-01
Feat: open files with relative path

singalhimanshu:
Now we can open files directly by running `amfora file.gmi` rather than
doing `amfora file:///absolute/path/to/file.gmi`

Solve #231

### DIFF
--- a/amfora.go
+++ b/amfora.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/makeworld-the-better-one/amfora/bookmarks"
 	"github.com/makeworld-the-better-one/amfora/client"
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one (changes requested):
Thanks for making this PR.

Checking for specific protocols as you have done will not work. For example, you have missed `https://`. But more importantly, the user could be using any protocol at all, thanks to the features Amfora has like [proxying](https://github.com/makeworld-the-better-one/amfora/wiki/Proxying).

A better solution is to check for `://`. If that's missing, then check if the file exists, just passing the argument as is. Absolute or relative paths can be handled the same, and so it doesn't need to be checked for that. If the file exists, convert to a `file://` URL and pass it along. If the file doesn't exist, pass it along unchanged, as it's maybe a domain like `geminispace.info`.

Let me know when you've made these changes (by comment or requesting a review), and I'll take a look again.

singalhimanshu:
Sorry for the delay I couldn't get any spare time. This might take some time if that is okay.

makeworld-the-better-one:
No worries. Just let me know if you can't continue with this later for whatever reason, and I can take it over.

makeworld-the-better-one:
Fixed by #257. Thanks for your work though!

PR #236, opened 2021-05-12
Added ability to save `about:` pages.

### LINKED ISSUE #210: Ability to save `about:` pages
I propose to implement ability to save subscription and bookmark page via CTRL + S. It could be use to publish our feeds and bookmark like CAPCOM. 

It would be interesting if Amfora could be run in command line (from for example crontab) save page, and we could automatically publish that saved pages to our Gemini capsule.

amohamed11:
Hi, I've been using Amfora for short-while and wanted to contribute to this awesome project.  
**Proposed Changes:**  
- Extracted the about page check that is in `hasContent` to its own separate function as that was the only check prohibiting about page saves. Since a page's content has no implication on it being an about page, decoupling these checks seems like a good idea. But do let me know if this change does not make sense for the overall project, and I can look into other fixes.    
- Added a change to `downloadNameFromURL` so that when it's generating a filename for a saved about page, it uses the entire page url i.e. `about:subscriptions` as otherwise the filename was simply left blank.

Fixes #210 
  
P.S. I did not locate a contribution guide, so let me know if I went about something the wrong way. If there is such a guide, please do point me towards it.

### DIFF
--- a/display/display.go
+++ b/display/display.go
@@ -155,7 +155,7 @@ func Init(version, commit, builtBy string) {
 				reset()
 				return
 			}
-			if query[0] == '.' && tabs[tab].hasContent() {
+			if query[0] == '.' && tabs[tab].hasContent() && !tabs[tab].isAnAboutPage() {
 				// Relative url
 				current, err := url.Parse(tabs[tab].page.URL)
 				if err != nil {
@@ -576,7 +576,7 @@ func Reload() {
 		return
 	}
 
-	if !tabs[curTab].hasContent() {
+	if !tabs[curTab].hasContent() || tabs[curTab].isAnAboutPage() {
 		return
 	}
 
--- a/display/display.go
+++ b/display/display.go
@@ -576,7 +576,7 @@ func Reload() {
 		return
 	}
 
-	if !tabs[curTab].hasContent() || tabs[curTab].isAnAboutPage() {
+	if !tabs[curTab].hasContent() {
 		return
 	}
 
--- a/display/download.go
+++ b/display/download.go
@@ -324,9 +324,7 @@ func downloadNameFromURL(dir, u, ext string) (string, error) {
 
 	parsed, _ := url.Parse(u)
 	if strings.HasPrefix(u, "about:") {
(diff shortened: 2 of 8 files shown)


display/display.go
```diff
@@ -576,7 +576,7 @@ func Reload() {
 		return
 	}
 
-	if !tabs[curTab].hasContent() {
+	if !tabs[curTab].hasContent() || tabs[curTab].isAnAboutPage() {
```
makeworld-the-better-one, line 579: Actually, reloading about pages should be allowed, so this can be removed.

display/download.go
```diff
@@ -321,8 +321,15 @@ func downloadPage(p *structs.Page) (string, error) {
 func downloadNameFromURL(dir, u, ext string) (string, error) {
 	var name string
 	var err error
+
 	parsed, _ := url.Parse(u)
-	if parsed.Path == "" || path.Base(parsed.Path) == "/" {
+	if strings.HasPrefix(parsed.String(), "about:") {
```
makeworld-the-better-one, line 326:
```suggestion
	if strings.HasPrefix(u, "about:") {
```

display/download.go
```diff
@@ -321,8 +321,15 @@ func downloadPage(p *structs.Page) (string, error) {
...
+		// Is an about page, use the entire page url since there is no hostname
+		name, err = getSafeDownloadName(dir, parsed.String()+ext, true, 0)
```
makeworld-the-better-one, line 328: I would just use the "path" or whatever. Like `about:subscriptions` should be saved as `subscriptions.gmi`.

makeworld-the-better-one (changes requested):
Thanks for making this PR!

There is no contribution guide at the moment, although I should definitely make one.

Overall this PR looks good! See my comments below for some small changes.

makeworld-the-better-one:
Thanks for the fix! Note I made a small change in 56b7e4a, because otherwise URLs like `about:subscriptions?2` would be saved as `subscriptions?2.gmi`, and `?` is a banned filename character on Windows.

Otherwise looks good, thanks for your work!

PR #239, opened 2021-05-15
Add iceberg theme

KNIX3:
I just thought it might be nice to add the iceberg theme found [here](https://cocopon.github.io/iceberg.vim/) to the list. Here's a screenshot. 
image


makeworld-the-better-one:
Thanks! Please add the screenshot to the `contrib/themes/README.md` file, as past PRs have done. See #221 as an example. It would also be nice if your screenshot has a left margin like the others do.

KNIX3:
I am more than willing to add such a screenshot, but what do you mean by a left margin? Is there a setting in amfora that adds a left margin or do you just want me to take the screenshot past the edge of the terminal?

makeworld-the-better-one:
By default Amfora adds a left margin to text, this can be seen in your screenshot. Strange if yours isn't, that could be a bug. This is controlled by the `left_margin` setting on the config.

KNIX3:
Ah, I found the setting, I don't know why mine was set to 0.0. I will change the image later today.

makeworld-the-better-one:
Thanks! Merged. Please subscribe to #149 so you can stay aware of updates that need to be added to this theme.

Issue #36, opened 2020-07-02
Search in pages


makew0rld:
<kbd>Ctrl-F</kbd> should bring up a bottom bar prompt to enter a search term. On <kbd>Enter</kbd> a `["search"]` region will be prepended to any bit of text that matches the search, with a `[""]` after it. Link selecting will have to be disabled, maybe with some ~~global~~ per-tab mode setting that switches between link, off, and search. <kbd>Tab</kbd> and <kbd>Backtab</kbd> can be used to navigate through all the found text fragments. <kbd>Esc</kbd> should switch the ~~global~~ per-tab mode back to off, and maybe use a regex to remove all the region open and close tags for the search.

makeworld-the-better-one:
The searching for the bit of text should ignore any style or region tags, so that may require a regex as well.

makeworld-the-better-one:
Having to add a per-tab mode setting is last straw of the code being somewhat disorganized. Using a `Tab` struct as mentioned in the NOTES.md would be helpful here.

makeworld-the-better-one:
The `tab` struct work is in progress on the `tab-struct` branch.

makeworld-the-better-one:
That work was completed.

lachrimae:
I'm investigating how to implement search in pages at the moment. Most of the difficulty seems to me to lie in the fact that it's difficult to reason about the state of the application in the relevant blocks of code. You can see, for instance, with [this tool](https://goreportcard.com/), that the number of `if/else` statements (they call it cyclomatic complexity) of the function `Init()` in amfora/display/display.go is very high. I have some ideas to simplify this that I want to run by you before I put a lot of work in. 

The point of this refactor is to make the organization of the code correspond to the state of the application. Code that should run under different application state should be in separate functions. When someone presses `Tab`, we want the resulting behaviour to depend on the application state, specifically, whether the page is in search mode or in normal mode. Right now, application state is managed inside `Init()` by some deeply-nested `if/else` and `case` statements. I want to test up-front for the mode of the tab and the page, and call separate functions according to the results of those tests. (In this block of code, the tab state is tracked with the variable `tabs[curTab].mode`, and the page state is tracked with `tabs[curTab].page.Mode`.)

In particular, I propose to
[...]

PR #240, opened 2021-05-30
adding tabModeSearch to search in page

Schubisu:
Addressing #36 I tried to follow up on that matter, because I really would love to have the 'search in page' feature implemented. I couldn't find the code that @lachrimae already worked on and I never really wrote a line of go before, so I expect that some of my edits can easily be improved. I'm thankful for any suggestions to make this PR better.

Instead of adding a "search" region tag, I add numbered tags to the matching strings, to make use of `cview`'s `ScrollToHighlight` function. I added a new tabMode to block other keypresses like follow links, as suggested in #36.

The way this is implemented, a search can potentially mess up other regions (e.g. if searched for special characters like `#` or `>`). I'm currently not sure how to address this best. I have not used regular expressions, not because I don't know how, but because I don't know how they can solve this problem. However, with an `Esc` keypress, the original buffer will be restored.


makeworld-the-better-one:
Hello, thanks for your contribution and interest in Amfora!

> Instead of adding a "search" region tag, I add numbered tags to the matching strings, to make use of cview's `ScrollToHighlight` function.

`ScrollToHighlight` scrolls so the highlighted regions are visible. But any region IDs can be made visible with `Highlight`, they don't need to be numbered strings.

> The way this is implemented, a search can potentially mess up other regions (e.g. if searched for special characters like `#` or `>`). I'm currently not sure how to address this best. I have not used regular expressions, not because I don't know how, but because I don't know how they can solve this problem.

Unfortunately it's pretty important to me that no page content gets messed up during searching. Regular expressions are needed because the content of the rendered version of the page can contain cview [color tags](https://docs.rocketnine.space/code.rocketnine.space/tslocum/cview#hdr-Colors) like `[red]` or `[red:white:]`. These tags are invisible, but change things like text color and style. If you searched for a string like "red" it would come up in these tags, but obviously the user does not want that result. And if you try to inject any region tags, it would mess up that tag entirely.

This is why searching for "#" causes issues, because a lot of the tags in your page look like `[#a3f4bb]` or something, and then trying to insert a numbered tag inside it breaks the color.

The solution to this is to use regexes to detect what's a tag and what's not, and skip over highlighting the things that are. This is all explained in #36. The steps needed to be implemented and the regex to use are all there.

If you'd like to update this PR to do that, I can leave it open, otherwise I'll close it. I appreciate you taking a stab at this, unfortunately it's trickier than any of us would like it to be.

Schubisu:
Hi @makeworld-the-better-one, thanks for your reply and suggestions.
Yes, I already thought that this would be the reason for not having a search implemented yet. Personally, to me a search function is more important than to leave the tags intact, given that the original buffer can be reloaded anytime with an `Esc` keypress. But I would like to try and make this PR better and implement the regex.

> `ScrollToHighlight` scrolls so the highlighted regions are visible. But any region IDs can be made visible with `Highlight`, they don't need to be numbered strings.

It's already some time ago that I looked into that, but I think that `ScrollToHighlight` does not work well if all region IDs are the same, because it will always scroll to the first region. That's why I numbered the regions to scroll to the next occurrence on tab press.

Please leave this PR open, so I can continue to work on this, and maybe others could also jump in. I think that I would have profited from the previous efforts that have been made in this direction, if the code had been visible here.

makeworld-the-better-one:
> But I would like to try and make this PR better and implement the regex.

Good to hear, thanks!

> It's already some time ago that I looked into that, but I think that `ScrollToHighlight` does not work well if all region IDs are the same, because it will always scroll to the first region. That's why I numbered the regions to scroll to the next occurrence on tab press.

Ah, you're right. Adding numbered region IDs will conflict with the link region IDs though, which are just numbers right now. Maybe for this the IDs could be like `search-0`, `search-1`, etc. That way it's still iterable.

Schubisu:
@makeworld-the-better-one  thanks again ;) I tried to implement all your suggestions; it seems to work well. The only flaw I currently see is that no search results are found inside of brackets that are no tags. Maybe that's an easy fix in the `tagsRegex` but I didn't find it, yet. Is it possible to switch between rendered and raw view inside amfora? That would be quite helpful.

makeworld-the-better-one:
Thanks for updating. I will review this PR in full soon. For now I came up with a new tags regex to try:
```regex
\[[a-zA-Z0-9_,;: \-\."#]+[^\[]*\]
```
Let me know how that works, and if it solves the issue or creates new ones.

> Is it possible to switch between rendered and raw view inside amfora? That would be quite helpful.

The raw view is stored inside each page struct. Not sure if you meant displaying the raw text, which would require escaping it then displaying it. There's no simple way to switch to displaying it built-in, but if you want to look at the raw page contents from the network, that's available in the struct.

Also there are some linter errors to look at.

config/config.go
```diff
@@ -250,6 +250,9 @@ func Init() error {
 	viper.SetDefault("keybindings.bind_copy_target_url", "c")
 	viper.SetDefault("keybindings.bind_beginning", []string{"Home", "g"})
 	viper.SetDefault("keybindings.bind_end", []string{"End", "G"})
+	viper.SetDefault("keybindings.bind_search", "/")
+	viper.SetDefault("keybindings.bind_next_match", "n")
+	viper.SetDefault("keybindings.bind_prev_match", "p")
```
makeworld-the-better-one, line 255: These keybindings need to be added to default-config.toml and to help.go. Don't forget to run `make gen` after doing that as well.

Also please change these keybindings to match Vim/less. See [this comment](https://github.com/makeworld-the-better-one/amfora/issues/36#issuecomment-746541058).

display/display.go
```diff
@@ -28,6 +28,12 @@ var termH int
 // The user input and URL display bar at the bottom
 var bottomBar = cview.NewInputField()
 
+var originalText []byte
+var searchBar = cview.NewInputField()
```
makeworld-the-better-one, line 32: Why a separate bar instead of using the existing one? Also this primitive would need to be themed like the bottom bar is. But I think just keeping one bar is better.

display/display.go
```diff
@@ -246,6 +252,89 @@ func Init(version, commit, builtBy string) {
...
+				originalText = tabs[curTab].view.GetBytes(false)
+			}
+			tabs[tab].mode = tabModeSearch
+
+			// find all positions of the search string
+			searchRegex := regexp.MustCompile(searchString)
+			searchIdx := searchRegex.FindAllIndex(originalText, -1)
+
+			// find all positions of tags
+			tagsRegex := regexp.MustCompile(`\[.*?[^\[]\]`)
```
makeworld-the-better-one, line 291: This regex should be changed as I noted in my previous comment, and also it can be compiled once globally, instead of in this function.

display/display.go
```diff
@@ -246,6 +252,89 @@ func Init(version, commit, builtBy string) {
...
+			lastMatch := 0
+			var isMatch bool
+
+			// loops through all occurrences and check if they
+			// discard if they lie within tags.
+			// []byte text is build from the original text buffer
+			// with the actual search strings replaced by tagged regions
+			// to highlight.
+			for i, match := range searchIdx {
+				isMatch = true
```
makeworld-the-better-one, line 305: Isn't this variable unnecessary? You already break out of the loop if it's false, so the check on 313 isn't needed either.

display/display.go
```diff
@@ -246,6 +252,89 @@ func Init(version, commit, builtBy string) {
...
+
+			// find all positions of the search string
+			searchRegex := regexp.MustCompile(searchString)
+			searchIdx := searchRegex.FindAllIndex(originalText, -1)
+
+			// find all positions of tags
+			tagsRegex := regexp.MustCompile(`\[.*?[^\[]\]`)
+			tagsIdx := tagsRegex.FindAllIndex(originalText, -1)
+
+			text := []byte("")
```
makeworld-the-better-one, line 294:
```suggestion
			text := make([]byte, 0)
```

display/display.go
```diff
@@ -246,6 +252,89 @@ func Init(version, commit, builtBy string) {
...
+					if match[0] >= tag[0] && match[1] <= tag[1] {
+						isMatch = false
+						break
+					}
+				}
+
+				if isMatch {
+					matches++
+					text = append(text, originalText[lastMatch:match[0]]...)
+					replacement := []byte(fmt.Sprint("[\"search-", i, "\"]", searchString, "[\"\"]"))
```
makeworld-the-better-one, line 316:
```suggestion
					replacement := []byte(fmt.Sprint(`["search-`, i, `"]`, searchString, `[""]`))
```
Using raw strings makes this much more readable.

display/display.go
```diff
@@ -291,6 +380,33 @@ func Init(version, commit, builtBy string) {
...
+				return nil
+			case config.CmdPrevMatch:
+				if curMatch > 0 {
+					curMatch--
+					tabs[curTab].view.Highlight(fmt.Sprint("search-", curMatch))
+				}
+				tabs[curTab].view.ScrollToHighlight()
+				return nil
+			case config.CmdInvalid:
+				if event.Key() == tcell.KeyEsc {
```
makeworld-the-better-one, line 400: You can ignore this case and just check the key with an if statement outside the switch.

makeworld-the-better-one (changes requested):
Thanks again for adding this! Super cool to see it in action. Please take a look at my comments above and below.

- Scroll position isn't saved after exiting search, so if I go back and go forward again it won't be in the right place
- Making two full copies of the page content feels very inefficient. I can't really think of a better way though, let me know if you have any ideas.
- Please replace `ScrollToHighlight` with `t.ScrollToHighlight`. This was just added to `master` so you'll have to update your branch here first.

hisacro:
can we have this as experimental feature? would be helpful even it is bugs

Schubisu:
@hisacro thanks, I completely forgot about this PR, but it's nice to you're interested ;)

@makew0rld my apologies for not following up on this earlier, and thank you so much for your thorough review. I think I have addressed all the points you mentioned, and tried to make small and clear commits for most of them. Please let me know if anything else is needed.

Schubisu (commented):
I think this should be fine this way, not sure about official review workflow in github. I guess I hit the "Submit review" button now. Please let me know if I can make any further improvements!

makew0rld:
Thanks. The diff on this doesn't look too bad, but in the near term I don't have time to test and add new features to Amfora. If you keep this PR open I might get to it eventually. You can see the readme for [details](https://github.com/makew0rld/amfora?tab=readme-ov-file#project-status).

makew0rld:
Thank you for this PR! It's working great.

Issue #205, opened 2021-02-27
Display gemtext file from stdin


mke21:
Is it possible for Amfora to read from stdin? I would like to pipe html2gmi directly into amfora if possible.

makeworld-the-better-one:
Amfora currently doesn't do this, but I can add it. For now your best bet would be to write to a file and then open that file like this: `amfora file:///absolute/path/to/file.gmi`.

Also if you're trying to browse the web with Amfora, I suggest you read [this](https://github.com/makeworld-the-better-one/amfora/wiki/Proxying) wiki page, and check out [Duckling Proxy](https://github.com/LukeEmmet/duckling-proxy)

mke21:
I'm actually trying to use this in combination with newsboat, an rss reader. Currently I use lynx after first piping through the readability tool. This works, but I much prefer gmi and the way amfora renders it. Setting up a proxy seems a bit overkill, but if that's the only way, then I will do that.. Thanks.

makeworld-the-better-one:
Well if you're writing some script I'd suggest writing the output to a temporary file, and then opening that file in Amfora on the next line. That's the only way to do it for now, but I will add this feature eventually.

dvejmz:
[I wrote a quick proof-of-concept](https://github.com/dvejmz/amfora/blob/%23205_stdin/amfora.go#L79) to see how this could work. Invoking this build of amfora with the following command renders the Gemtext file supplied via stdin to a new tab.

```
sample-gemtext.gmi  | ./amfora 
```

I'd be happy to iterate on this and properly integrate this feature with the rest of the application.

makeworld-the-better-one:
Closed by #242.

PR #242, opened 2021-06-11
Render stdin text as Gemtext (#205)

dvejmz:
This allows `amfora` to accept content via stdin and render it as Gemtext. The main use case this fulfils is the rendering of Gemtext files via stdin (see #205) but plaintext is also handled.

### DIFF
--- a/amfora.go
+++ b/amfora.go
@@ -94,7 +94,5 @@ func renderFromStdin() {
 	}
 
 	stdinText := stdinTextBuilder.String()
-	if len(strings.TrimSpace(stdinText)) > 0 {
-		display.RenderFromString(stdinText)
-	}
+	display.RenderFromString(stdinText)
 }
--- a/amfora.go
+++ b/amfora.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/makeworld-the-better-one/amfora/bookmarks"
 	"github.com/makeworld-the-better-one/amfora/client"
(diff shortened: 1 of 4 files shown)


amfora.go
```diff
@@ -68,10 +70,34 @@ func main() {
...
 
 	// Start
 	if err = display.App.Run(); err != nil {
 		panic(err)
 	}
 }
+
+func isStdinEmpty() bool {
+	stat, _ := os.Stdin.Stat()
+	return (stat.Mode() & os.ModeCharDevice) != 0
```
makeworld-the-better-one, line 85: Is this check portable across OSes? Where did you find this?

amfora.go
```diff
@@ -68,10 +70,34 @@ func main() {
...
+}
+
+func renderFromStdin() {
```
makeworld-the-better-one, line 88: Why not just use `io.Copy` here? It would be a lot more efficient. And then keep it as bytes until the last moment you need to change it, in `renderPageFromString`.

makeworld-the-better-one (approved):
Thanks for creating this PR! Works well for me, just see my comments below.

amfora.go
dvejmz: This check leverages [the Go standard library](https://golang.org/pkg/io/fs/#ModeCharDevice) so I'd have thought it would be. I haven't tested this outside of my Fedora Linux installation so I wouldn't be able to say for sure though.

I got the idea from [here](https://stackoverflow.com/a/26567513/2115499). [This article](https://flaviocopes.com/go-shell-pipes/) explains the approach in a bit more detail.

makeworld-the-better-one: Seems good. [This answer](https://stackoverflow.com/a/41999124/) mentions the same solution working on macOS too. And I guess pipes don't really apply to Windows.

amfora.go
```diff
@@ -68,10 +70,31 @@ func main() {
...
+	stdinTextBuilder := new(strings.Builder)
+	_, err := io.Copy(stdinTextBuilder, os.Stdin)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error reading from standard input: %v\n", err)
+		os.Exit(1)
+	}
+
+	stdinText := stdinTextBuilder.String()
+	if len(strings.TrimSpace(stdinText)) > 0 {
```
makeworld-the-better-one, line 97: I don't think this check is necessary. Nothing wrong with displaying some spaces or just nothing at all. Would be useful if part of some longer pipeline, where returning nothing is possible but not certain. Much better to get an empty page than be confused by the new tab page.

Other than this I think we're good to go!

makeworld-the-better-one:
Thanks for adding this!

Issue #181, opened 2021-02-06
Using the user's terminal colors


Zambito1:
Thank you for making Amfora, it is quite a nice browser!

I saw you mention that [you were having trouble using the users terminal colors](https://github.com/makeworld-the-better-one/amfora/issues/46#issuecomment-665279904). It seems that when launching the terminal over SSH, it properly uses the users terminal colors. Maybe this can uncover some way that it can be done an easier way.

Both terminals are here are Alacritty, using the Nord colors. The left is using ssh to open Amfora, the right is not.
![Screenshot of two amfora windows, the left using the terminals colors, the right using default black with high contrast](https://user-images.githubusercontent.com/7004857/107102383-5e791800-67e8-11eb-8102-9817b24164d3.png)

makeworld-the-better-one:
I believe this how to do with how tcell (low-level terminal library, used by Amfora and the TUI middle-layer library called cview) works. In a normal terminal it can check for truecolor support and attempt to display the original color. Over SSH, tcell thinks truecolor isn't supported, and it will fallback to the regular color codes, which the terminal will then display using the built-in palette. That's my guess, anyway.

IIRC the reason Amfora doesn't try to use default terminal colors is that it's hard to get both light and dark terminals to work correctly. There's not really a way to know whether to use white text or black text for example, and I believe that just trying to use terminal colors will mess up in certain situations, like put white text on the default white terminal background. I might be wrong about that, and it's worth looking into. But it's easier to just use truecolor and provide support for themeing. For example, there is a [Nord theme](https://github.com/makeworld-the-better-one/amfora/tree/master/contrib/themes#nord). But I understand that's annoying to create for other terminal setups.

I leave this open as a question for now, something that should be tested more. Currently I am updating tcell in #107, and right now I've set things up so truecolor is forced whenever possible, ignoring the user's terminal colors.

objectliteral:
I second the request for respecting the terminal's color scheme, but I agree that doing so is not without some limitations.

> But it's easier to just use truecolor and provide support for themeing.

It's slightly annoying for users to go to each application's config and tweak the color scheme to match the rest of their systems. Especially so if they have set up some kind of automation like with pywal.

> IIRC the reason Amfora doesn't try to use default terminal colors is that it's hard to get both light and dark terminals to work correctly. There's not really a way to know whether to use white text or black text for example, and I believe that just trying to use terminal colors will mess up in certain situations, like put white text on the default white terminal background.

[...]

PR #245, opened 2021-06-20
Fix issue 244

m-mcdonagh:
Added support for Terminal Default Colors (which may or may not be transparent)

- Specifying `"none"` in your config file sets the color to the default color of the terminal, or `tcell.ColorDefault`, as defined [here](https://pkg.go.dev/github.com/gdamore/tcell#ColorDefault).
- Invalid colors still result in an error, so malformed config files alert the user instead of setting the color to the default.
- Updated the default config and color string for [cview color tags](https://docs.rocketnine.space/code.rocketnine.space/tslocum/cview#hdr-Colors) to reflect this

On terminals which support transparency and are configured as such, setting `bg = "none"` keeps the transparency

example with the simple terminal using the alpha patch:
example


config/config.go
```diff
@@ -349,9 +349,15 @@ func Init() error {
 			if !ok {
 				return fmt.Errorf(`value for "%s" is not a string: %v`, k, v)
 			}
-			color := tcell.GetColor(strings.ToLower(colorStr))
-			if color == tcell.ColorDefault {
-				return fmt.Errorf(`invalid color format for "%s": %s`, k, colorStr)
+			colorStr = strings.ToLower(colorStr)
+			var color tcell.Color
+			if colorStr == "none" {
```
makeworld-the-better-one, line 354:
```suggestion
			if colorStr == "none" && k == "bg" {
				// Getting the terminal default only works for the background right now
```

makeworld-the-better-one (changes requested):
Nice solution, thank you. And it looks good in your screenshot! I've never seen Amfora like that before, it's the right way to go.

Since your solution only works for backgrounds (for text it just seems to use white every time) I've suggested a change that will make it only apply when used with `bg`. Please note that in `default-config.toml`.

Also, this should be changed from "none" to "default", I think that name works better.

After making your changes, run `make gen` (or just `go generate ./...`), like I did for this PR earlier in df21b0b. This will copy the config change into the code.

m-mcdonagh:
Sure thing! I'll get right to working on that.
Are you sure you want to limit it just to just `bg`? I think the user could decide what to make the default

Maybe, limit `default` only to values which affect the `<background>` of [color tags](https://docs.rocketnine.space/code.rocketnine.space/tslocum/cview#Box.SetBackgroundTransparent#hdr-Colors) `[<foreground>:<background>:<flags>]`, which I'm assuming is all the settings ending in `bg`, but I'll double check before finalizing the PR, if this is the way you'd like to go.

Also, no need to worry about the "transparency". It will always show right through the terminal instead of revealing any widgets behind, because the background isn't really transparent, it's "filled" with the color, and overwrites the pixels there. Transparency to reveal other widgets would require [this](https://docs.rocketnine.space/code.rocketnine.space/tslocum/cview#Box.SetBackgroundTransparent)

Here is an example where both the `bg` and `info_modal_bg` are set to `"none"`, soon to be `"default"`. (I also set `info_modal_text = "slateblue"` to make it more readable)

makeworld-the-better-one:
> Maybe, limit `default` only to values which affect the `<background>` of color tags

Sounds good to me! Good point that those ones could make use of this setting as well.

m-mcdonagh:
@makeworld-the-better-one 
I completed what you asked of me. 
I also found a UI quirk:
When a UI element was selected, the bg and text would swap. However, the if bg was `"default"`, which could be dark, now the text was `"default"`, which is always white. This lead to light text on a light background. I circumvent this with a new function `GetTextColor(key, bg string) tcell.Color`, which, can return black or white depending on the brightness of bg, only if the key is ColorDefault. Otherwise it just returns the color in the theme map
pic-full-210629-180837
pic-full-210629-180845

Also, `"default"` is valid for all theme settings ending in "bg", and if the user tries setting anything else to `"default"`, they get the following error: 
pic-selected-210629-1758-02

makeworld-the-better-one:
Thanks for updating!

> the text was `"default"`, which is always white. This lead to light text on a light background. This lead to light text on a light background. I circumvent this with a new function

I mentioned that in my previous [comment](https://github.com/makeworld-the-better-one/amfora/pull/245#pullrequestreview-692216390). I don't think this function should happen. Until using `default` for text can work, whatever the user specified for text color (even white on white) should be respected.

m-mcdonagh:
Yes, so the user cannot specify `"default"` for text, but because they can specify UI elements to have a background of `"default"`, and the text and bg switch colors on focus, text can become ColorDefault on focus.

My code always respects the user's config. The only time the text can change depending on the background is when a UI element is focused, and its background is `"default"`. The effect of my code is the same as yours otherwise

config/theme.go
```diff
@@ -84,5 +84,44 @@ func GetColor(key string) tcell.Color {
...
+
+// GetReadableColor Returns ColorBlack if the color is brighter than gray
+// otherwise returns ColorWhite
+func GetReadableColor(bg tcell.Color) tcell.Color {
+	if bg == tcell.ColorDefault {
+		return tcell.ColorWhite
+	}
+	r, g, b := bg.RGB()
+	const gray = 0x88 + 0x88 + 0x88 // The sum of gray's R, G, and B
+	if r+g+b > gray {
```
makeworld-the-better-one, line 102: The "middle gray" as perceived by humans is 119. So I would set `gray` equal to that. I would calculate the gray value of the color more accurately as `77*r+150*g+29*b + (1<<7) >> 8`, and then compare that to `gray`. Overall this will be more accurate in determining color luminance.

Otherwise looks good, thank you!

config/theme.go
```diff
@@ -84,5 +84,45 @@ func GetColor(key string) tcell.Color {
...
+		return "-"
+	}
+	return fmt.Sprintf("#%06x", color.Hex())
+}
+
+// GetReadableColor Returns ColorBlack if the color is brighter than gray
+// otherwise returns ColorWhite
+func GetReadableColor(bg tcell.Color) tcell.Color {
+	if bg == tcell.ColorDefault {
+		return tcell.ColorWhite
```
makeworld-the-better-one, line 98: Doesn't this assume that the default background is dark, so white text would be a good choice? This comes to the main problem with #181, which is knowing the user's terminal (and text) color. Once that is solved (and it can be, I just haven't implemented it yet), everything else can be done.

config/theme.go
```diff
@@ -84,5 +84,45 @@ func GetColor(key string) tcell.Color {
...
+	}
+	r, g, b := bg.RGB()
+	luminance := 77*r + 150*g + 29*b + (1<<7)>>8
```
makeworld-the-better-one, line 101:
```suggestion
	luminance := (77*r + 150*g + 29*b + 1<<7) >> 8
```
Gave you the wrong math before, sorry.

config/theme.go
```diff
@@ -84,5 +84,45 @@ func GetColor(key string) tcell.Color {
...
+	const gray = 0x77 + 0x77 + 0x77 // The sum of gray's R, G, and B
```
makeworld-the-better-one, line 102:
```suggestion
	const gray = 119 // The middle gray
```

It doesn't need to be added 3 times because `luminance` is not the sum of RGB, it's one value from 0-255 that represents them all.

display/bookmarks.go
```diff
@@ -41,8 +41,10 @@ func bkmkInit() {
 		form.SetLabelColor(config.GetColor("bkmk_modal_label"))
 		form.SetFieldBackgroundColor(config.GetColor("bkmk_modal_field_bg"))
 		form.SetFieldTextColor(config.GetColor("bkmk_modal_field_text"))
+		form.SetFieldBackgroundColorFocused(config.GetColor("bkmk_modal_field_text"))
+		form.SetFieldTextColorFocused(config.GetTextColor("bkmk_modal_field_bg", "bkmk_modal_field_text"))
 		form.SetButtonBackgroundColorFocused(config.GetColor("btn_text"))
-		form.SetButtonTextColorFocused(config.GetColor("btn_bg"))
+		form.SetButtonTextColorFocused(config.GetTextColor("btn_bg", "btn_text"))
```
makeworld-the-better-one, line 47: Isn't this function call, and all the others in this PR, incorrect? The function signature for `GetTextColor` is `GetTextColor(key, bg string) tcell.Color`, which would indicate having `"btn_bg"` second and `"btn_text"` first. Or am I misunderstanding?

makeworld-the-better-one (changes requested):
Sorry to keep going back and forth on this, just noticed some more things as I went through the PR. Thanks for your work.

display/bookmarks.go
m-mcdonagh: for focused UI elements, the text color and bg color swap. When focused, the button has text color of whatever `"btn_bg"` is set to, and a bg of whatever color `"btn_text"` is set to.

Because `"default"` is not a valid text color in the config, this function is always called with a key ending with `bg` as its key, and a `bg` ending with text. 

I tried to be as concise as possible with the naming. Maybe `SwapColor(new_text_key, new_bg_key string)` or something similar would be less confusing.

config/theme.go
m-mcdonagh: yes it does, but because we reject any text fields with a value of `"default"`, and the `bg` argument is always set to a string ending in `"text"`, the if condition is never true. I just added that because `tcell.ColorDefault.RGB()` returns `-1, -1, -1`, so if this function was used in the future, I didn't want unexpected calculations to run if tcell.ColorDefault was passed in.

Now that I think about it, a better solution might be to return `tcell.ColorDefault`, but I'll let you decide

makeworld-the-better-one: Oh, I see. Could you rename `bg` to be more appropriate, seeing as it's not backgrounds, but like text colors, or rather the backgrounds being set to the color of the text? To be honest maybe the function should be renamed to `GetBlackWhite`, and `bg` to `c`, with comments that explain what it does, and the case for `tcell.ColorDefault`.

> Now that I think about it, a better solution might be to return `tcell.ColorDefault`, but I'll let you decide

I think that makes more sense. Please also add a comment  in the function explaining why this shouldn't happen anyway.

display/bookmarks.go
makeworld-the-better-one: No, I think it's good the way it is then, I just misunderstood.

makeworld-the-better-one:
Thank you for working on this! Sorry for sleeping on it, don't be afraid to ping if you need in the future.

m-mcdonagh:
happy to help :)

PR #253, opened 2021-09-11
🧛🏻‍♂️ Add Dracula theme variant

marcransome:
I've added a variant of the Dracula theme that more closely aligns with the [Dracula Specification](https://spec.draculatheme.com). Let me know what you think! I'm happy to maintain this and have subscribed to https://github.com/makeworld-the-better-one/amfora/issues/149 to stay up to date with future changes.


makeworld-the-better-one:
Looks good, thanks for adding it!

PR #257, opened 2021-09-28
Allow opening local files by relative path

### LINKED ISSUE #231: Open files with relative path

justjosias:
Local files can now be loaded by relative path, rather than the
whole path preceded by the `file://` scheme indicator. It first
checks whether or not the requested URL has a scheme (by checking
for `://` in the URL). If not, it checks to see if the file exists
before opening. It also allows making intention known by
specifying `./` or `../` before the URL, closely matching the
behavior of Firefox.

This borrows code from @singalhimanshu's PR (#232).

Resolves #231

### DIFF
--- a/amfora.go
+++ b/amfora.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/makeworld-the-better-one/amfora/bookmarks"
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
Looks good, thank you!

PR #258, opened 2021-09-29
add omni by day & by night themes

omnivagant:
I think the commented out values are the default ones, but it was some months since I did this. I could probably clean this up a bit, but I wanted to share.

Myself, I now use transparent terminals and background = "default".

### DIFF
--- a/contrib/themes/README.md
+++ b/contrib/themes/README.md
@@ -157,6 +157,18 @@ Contributed by **[@luetage](https://github.com/luetage)**
 
 ![screenshot of Tokyo Night theme](https://user-images.githubusercontent.com/13988217/130348393-69986b51-ddd7-4310-90ae-382461502535.png)
 
+## omni by day
+
+Contributed by **[@omnivagant](https://github.com/omnivagant)**
+
+![screenshot of omni-by-day v0.23](https://user-image.githubusercontent.com/omnivagant/cant-figure-out-how-to-do-this.png)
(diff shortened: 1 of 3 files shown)


makeworld-the-better-one:
Thanks for sharing. If you're able to clean this up that would be appreciated, you can see other themes as a reference. It would make maintenance much easier in the future.

makeworld-the-better-one:
@omnivagant are you planning on updating this? Changes will be required now since the Amfora default theme has changed. I suspect it will mostly just be uncommenting all your lines.

You also need to update the theme README with some information and a photo, see #251 or any other theme PR as an example.

makeworld-the-better-one:
@omnivagant looking at your latest commits:

To upload images, you can put them in this thread, then get the link from there.

For `bg`, just put the value that best fits your theme. Presumably it's `#ffffff` for day and `#000000` for  night.

Also, your theme is missing keys. Please define all the possible theme keys, as Amfora (v1.9.0 and later) will default to using the 
terminal theme for undefined keys.

For the future I prefer PR authors to not do force-pushes as it makes it hard to understand what was changed. I will squash the commits when merging anyway.

Thanks!

makeworld-the-better-one:
@omnivagant any update on this?

makeworld-the-better-one:
I'm going to close this, but feel free to open another PR taking my comments above into account.

PR #259, opened 2021-10-03
Add RPM spec file

elagost:
I made a basic spec file for packaging in RPM-based distributions. When new versions of amfora are created the spec file should be updated at the top (`Version: 1.8.0` currently) and the bottom (new entry under `%changelog`).
It builds in current Fedora versions, RHEL8, and openSUSE Tumbleweed. https://copr.fedorainfracloud.org/coprs/adamthiede/bin/build/2872181/

### DIFF
--- a/amfora.spec
+++ b/amfora.spec
@@ -1,0 +1,44 @@
+Summary: a Gemini browser for your terminal
+Name: amfora
+Version: 1.8.0
+Release: 1%{?dist}
+License: GPL
+URL: https://github.com/makeworld-the-better-one/amfora
+Source: https://github.com/makeworld-the-better-one/amfora/archive/v%{version}.tar.gz
+
+BuildRequires: make
+BuildRequires: git
+BuildRequires: gcc
+BuildRequires: golang
+
+Requires: ncurses-base
+
+%global debug_package %{nil}
+
+%description
+Amfora aims to be the best looking Gemini client with the most features... all in the terminal.
+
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
Thanks for adding this. Is this needed to get Amfora into Fedora then? I assumed distro maintainers could create this instead of it needing to be in the repo, but I'm fine if it needs to be.

You should set the `BUILDER` env var when calling `make`. Otherwise it would default to a value of `Makefile`, when it should be something like `official-rpm`, or `official-fedora`.

elagost:
This file just makes it easier for individuals to build, and for the package to be adopted into distributions. Distro maintainers would be able to do it themselves. I'm not a contributor to any distro (yet, I hope to change that) but just thought this would be a helpful addition.
If you'd prefer me to go through the packaging process for Fedora, I will make an attempt. Something I've been wanting to do for a while honestly. I'll define the `BUILDER` variable as well.

makeworld-the-better-one:
The ideal situation for me is when distro package maintainers can handle any distro-specific information or files, and so that way I don't need to update it. So yeah, it'd be great if you could go ahead and make the Fedora package instead. Does that work?

elagost:
Sounds good. I will try and make that happen. Thank you for being so communicative.

makeworld-the-better-one:
You're welcome, and thank you for putting time into this! Let me know what happens.

PR #260, opened 2021-10-09
Switch on debug logs via env var flag

dvejmz:
Use an environment variable as a flag to easily switch on the amfora
debug logger without requiring a code change.

Sample usage

```
$ DEBUG=1 amfora
```

The rest of the logging behaviour has been preserved although I'm thinking of further improvements that could be made to it, such as structured logs for easier filtering and parsing by tools (e.g. JSON and jq), increase log coverage, etc.

### DIFF
--- a/amfora.go
+++ b/amfora.go
@@ -21,14 +21,14 @@ var (
 )
 
 func main() {
-	debugModeEnabled := os.Getenv("DEBUG")
-	if debugModeEnabled != "" {
-		err := logger.Init()
-		if err != nil {
-			panic(err)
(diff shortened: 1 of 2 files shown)


makeworld-the-better-one:
This looks good, thanks! I only really use the log for debugging, and then I remove all calls. But this will make it easier, and I'm open to there being logging calls left in there if anyone would like to add them.

Note that in 8e6ca6e I changed the env var from `DEBUG` to `AMFORA_DEBUG` to prevent collisions with other applications.

PR #261, opened 2021-10-09
Fix users being unable to exit help menu

### LINKED ISSUE #241: Cannot exit help menu after editing the URL or changing the tab
When I open the help page using `?` and then I edit the current URL using `e` and press Enter, I am no longer able to exit the help page using `Esc`. 

This issue also seems to also be present when I open the help page and then try to open a new tab using `Ctrl+T`, or move to a different tab using `F1` or `F2`, or reload the page using `R`.

In this situation, the only way (which I know of) to make the help menu go away is to completely exit Amfora.

Is this a bug, or is there something which I'm misunderstanding about the keyboard controls?

dvejmz:
Ignore any keypresses that don't have a corresponding binding in the Help page to avoid users triggering keybindings that remove the focus from the Help screen and becoming unable to exit, as Help keybindings are no longer captured correctly.

Rather than adding unnecessary complexity to make the Help page more flexible and capable of handling input events meant for other panels, I opted to restrict the input set to the keybindings that have a corresponding action in the Help page while it's active.

Closes #241 .

### DIFF
--- a/display/bookmarks.go
+++ b/display/bookmarks.go
@@ -29,7 +29,7 @@ var bkmkCh = make(chan bkmkAction)
 var bkmkModalText string // The current text of the input field in the modal
 
 func bkmkInit() {
-	panels.AddPanel("bkmk", bkmkModal, false, false)
+	panels.AddPanel(PanelBookmarks, bkmkModal, false, false)
 
 	m := bkmkModal
 	if viper.GetBool("a-general.color") {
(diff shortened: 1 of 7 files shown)


makeworld-the-better-one:
Looks like an elegant solution, thanks. Using those constants is nice too. However, this removes some keybindings that are probably important to keep, like the ones for quitting Amfora. Ideally those would still work.

I know it's been a while, sorry about missing the Hacktober fest thing. Let me know if you'd like to continue working on this PR, and I'll take over if not.

makeworld-the-better-one:
I've made some changes so that the quit key is supported. Everything else should be fine. Thanks for your PR!

Issue #252, opened 2021-08-22
Syntax highlighting for preformatted text blocks that have alt text indicating a programming language


mntn-xyz:
From the Gemini spec:

"Anything which comes after the ``` characters of a line which toggles preformatted line *on* (i.e. the first, third, fifth, etc. toggling lines in a document) may be treated as "alt text" for the preformatted content. In general you should not count on this content being visible to the user but, for example, search engines may index it and screen readers may read it to users to help the user decide whether the preformatted content should be read aloud (which e.g. ASCII art generally should not be, but which source code perhaps should be). There are currently no established conventions on how alt text should be formatted."

Many Markdown implementations, for instance, use a similar mechanism to identify the language for syntax highlighting:

~~~
```html
<p>This is HTML</p>
```
~~~

It seems like it is inevitable to have some kind of syntax highlighting convention in Gemini, especially given its usage within tech-centric circles. Given that the spec leaves alt text wide open for interpretation by the client, I thought that it might be worth investigating this use case in amfora.

Hugo uses a fantastic Go library called [Chroma](https://github.com/alecthomas/chroma) for its syntax highlighting. Chroma explicitly supports highlighting with 8-bit and 256-bit terminal colors, so it should be a good match.

mntn-xyz:
I have a patch almost ready for this, see https://github.com/mntn-xyz/amfora/tree/chroma. It can be tested on this page: gemini://mntn.xyz/posts/2021-08-22-hello-world/

There are two problems:
1. On one very long preformatted text line, it seems like a couple of characters are missing their formatting. I'm not sure if this is a bug with chroma's terminal formatter or something with amfora.
2. For some reason every preformatted block has an extra newline added after it if it uses syntax highlighting.

Anyway, I dropped this here for now in case someone else wants to take a look. I'll pick it up again soon.

mntn-xyz:
Here's a screenshot too:

![image](https://user-images.githubusercontent.com/85877297/131273272-6b4c3b21-248d-4eae-ae2f-2a0eb527c674.png)

makeworld-the-better-one:
Thanks for creating this issue, it's an interesting idea. I do kind of hesitate to add things like this because they're not standardized, but also the meaning of an alt text that is just `python` or something is pretty clearly for syntax highlighting. I would be interested in seeing any discussion of this on the mailing list and real world use if you have any offhand.

[...]

PR #263, opened 2021-10-12
Add Chroma support for syntax highlighting

mntn-xyz:
For #252.


mntn-xyz:
Re: the shifted colors, I'm starting to wonder if it has to do with the close bracket from Markdown links somehow confusing things. It's not a problem with any other highlighted code, and it only happens after that close bracket.

mntn-xyz:
TODO:

- [ ] Allow theme selection in config file
- [ ] Allow disabling of syntax highlighting in config file
- [ ] Detect terminal colorspace and use that (possibly; 8 colors are probably enough)
- [ ] Add documentation to README, wiki, default config file

default-config.toml
```diff
@@ -44,6 +44,12 @@ color = true
 # Whether ANSI color codes from the page content should be rendered
 ansi = true
 
+# Whether or not to support source code highlighting in preformatted blocks based on alt text
+highlight_code = true
+
+# Which highlighting style to use (see https://xyproto.github.io/splash/docs/)
+highlight_style = "swapoff"
```
makeworld-the-better-one, line 51: I find this theme pretty ugly and a poor use of color. Could it be switched to "vim"? Make sure to change the default everywhere.

renderer/renderer.go
```diff
@@ -289,11 +302,46 @@ func RenderGemini(s string, width int, proxied bool) (string, []string) {
...
+			if formatter == nil {
+				formatter = formatters.Fallback
+			}
+			lexer := lexers.Get(lang)
+			if lexer == nil {
+				lexer = lexers.Fallback
+			}
+
+			// Strip existing ANSI codes
+			strippedBuf := ansiRegex.ReplaceAllString(buf, "")
```
makeworld-the-better-one, line 329: Can't `buf` be reused for this, reducing memory usage?

makeworld-the-better-one (changes requested):
Thanks for your PR! This is a nice feature. Please see my comments below, and note there is a merge conflict to be resolved.

makeworld-the-better-one:
> Updated Wiki documentation
> 
> [Home.md](https://github.com/makeworld-the-better-one/amfora/files/7350783/Home.md)
> 
> [Source-Code-Highlighting.md](https://github.com/makeworld-the-better-one/amfora/files/7350784/Source-Code-Highlighting.md)

What changes did you make to the homepage? It looks the same to me. Thanks for the highlighting page though, I'll use it for the wiki when this is merged.

mntn-xyz:
Thanks for reviewing, I'll get the changes made ASAP. Re: the wiki, I meant to include an updated sidebar page, good catch:

[_Sidebar.md](https://github.com/makeworld-the-better-one/amfora/files/7673072/_Sidebar.md)

makeworld-the-better-one:
Thanks! About to release v1.9.0, but this will make it into the next minor version.

renderer/renderer.go
mntn-xyz: The ANSI codes are only stripped if there is no error formatting the text using the detected language; if there is an error we just display any existing ANSI codes and treat it as normal preformatted text. That's why I was keeping buf around.

To possibly reduce memory usage I can remove 329 and change 332 to the following:
```
iterator, err := lexer.Tokenise(nil, ansiRegex.ReplaceAllString(buf, ""))
```

I thought that maybe the go compiler would optimize out strippedBuf, but after a quick search it sounds like some people complain that go sometimes doesn't do enough optimization. This should be a simple way to ensure that we avoid creating extra duplicates of buf.

default-config.toml
mntn-xyz: No problem, swapoff looked good with the old default theme but vim looks better now!

README.md
```diff
@@ -162,6 +163,8 @@ Amfora ❤️ open source!
 - [progressbar](https://github.com/schollz/progressbar)
 - [go-humanize](https://github.com/dustin/go-humanize)
 - [gofeed](https://github.com/mmcdole/gofeed)
+- [chroma](https://github.com/alecthomas/chroma) for source code syntax highlighting
+- [go-supportscolor](https://github.com/jwalton/go-supportscolor)
```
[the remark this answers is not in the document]
mntn-xyz: Whoops, yes termenv looks like a better option!

renderer/renderer.go
```diff
@@ -5,20 +5,33 @@
...
 )
 
+// Terminal color information, set during display initialization by display/display.go
+var TermColor string
+
 // Regex for identifying ANSI color codes
 var ansiRegex = regexp.MustCompile(`\x1b\[[0-9;]*m`)
 
+// Regex for identifying possible language string, based on RFC 6838 and lexers used by Chroma
+var langRegex = regexp.MustCompile(`^([a-zA-Z0-9]+/)?[a-zA-Z0-9]+([a-zA-Z0-9!_\#\$\&\-\^\.\+]+)*`)
```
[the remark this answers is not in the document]
mntn-xyz: Generally what the regex handles is textAndNumbers/textAndNumbersWithOptionalSymbolsAfter, which matches all known language MIME types that I could find. I wanted to limit the match as much as possible to allow people to write things like "text/javascript: sample code" and still be detected correctly.

I looked at what MIME type names are possible as defined in RFC 6838 and then expanded that based on the MIME types actually supported in Chroma (some of which are not compliant with RFC 6838). It is possible that Chroma may introduce a formatter for some new niche language with a MIME type like text/COOL**TROLL**LANGUAGE, but this seems unlikely. The worst that may happen is that the regex doesn't detect an obscure new language that is introduced; generally this regex covers all common language names and then some.

mntn-xyz:
Everything should be resolved now.

renderer/renderer.go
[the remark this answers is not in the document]
makeworld-the-better-one: Thanks for explaining. It's nice that Chroma can handle MIME types. I see that this will capture non-MIME-type strings as well, great.

display/display.go
```diff
@@ -60,6 +61,18 @@ var App = cview.NewApplication()
...
+	// Detect terminal colors for syntax highlighting
+	switch termenv.ColorProfile() {
+	case termenv.TrueColor:
+		renderer.TermColor = "terminal16m"
+	case termenv.ANSI256:
+		renderer.TermColor = "terminal256"
+	case termenv.ANSI:
+		renderer.TermColor = "terminal16"
+	default:
+		renderer.TermColor = "terminal8"
```
makeworld-the-better-one, line 73: Code highlighting should not be happening at all in this case, as this means the terminal does not support ANSI at all.

renderer/renderer.go
```diff
@@ -315,11 +328,43 @@ func RenderGemini(s string, width int, proxied bool) (string, []string) {
...
+					// Strip extra newline added by Chroma and replace buffer
+					buf = string(trailingNewline.ReplaceAll(formattedBuffer.Bytes(), []byte{}))
+				}
+			}
+		}
+
 		// Support ANSI color codes in preformatted blocks - see #59
-		if viper.GetBool("a-general.color") && viper.GetBool("a-general.ansi") {
+		// This will also execute if code highlighting language is set
+		if viper.GetBool("a-general.color") && (viper.GetBool("a-general.ansi") || lang != "") {
```
makeworld-the-better-one, line 367: There needs to be multiple if statements or something here, to support a configuration where the user doesn't want ANSI rendered, but they do want code highlighted, or vice versa.

default-config.toml
```diff
@@ -55,6 +55,12 @@ color = true
 # Whether ANSI color codes from the page content should be rendered
 ansi = true
 
+# Whether or not to support source code highlighting in preformatted blocks based on alt text
+highlight_code = true
+
+# Which highlighting style to use (see https://xyproto.github.io/splash/docs/)
+highlight_style = "vim"
```
makeworld-the-better-one, line 62: Upon some testing, I find the vim is pretty difficult to read, especially that blue for comments. I find that monokai is a vast improvement and a good default. Let me know if you disagree, but otherwise I'd like to switch it.

makeworld-the-better-one (changes requested):
Thanks for making changes! Spotted some more things after looking again. Also note that lint CI is failing.

default-config.toml
mntn-xyz: Sure, I like monokai and it's a good default. Source highlighting themes are intensely personal!

display/display.go
mntn-xyz: You're right, it's just that there's a gap since termenv can't detect 8-color terminals. On the other hand, I'm not sure how many of those exist anyway.

renderer/renderer.go
mntn-xyz: OK, that makes sense, I was thinking of it more from the perspective of "highlighting uses ANSI, so using highlighting requires ANSI" but it probably does make more sense to think of them as two separate features... people may want code highlighting without arbitrary colorful preformatted text.

mntn-xyz: Ok, actually this is almost correct. So this only executes on a block if ANSI highlighting is on OR if a language is detected, but instead it should execute if ANSI highlighting is on OR if the syntax was *successfully tokenized* based on the detected language. This requires a new boolean variable that is set if err == nil from the call to lexer.Tokenise().

mntn-xyz:
Should be ready to go again!

renderer/renderer.go
makeworld-the-better-one: I was misunderstanding before I think, but thanks for catching that bug as well. Looks good now.

makeworld-the-better-one:
Looks good now! Thanks a lot for this work. I've added your pages to wiki as well with some minor edits, take a look and let me know here if you have any suggestions.

Issue #262, opened 2021-10-11
Ability to run user-defined actions/scripts


mntn-xyz:
It would be nice to have a hotkey or bar command to run a user-defined action with the argument being the current URL or the URL of a specific link #. For instance, I might want to append the URL to a text file, use Emacs org-capture to file the URL appropriately, compose a new email with the link in the message body, etc.

mntn-xyz:
True, I wouldn't consider it high priority, but it would save a couple of keystrokes each time, which can add up.

What led me to suggest this is that I like to manage bookmarks in orgmode, as I can annotate, categorize, and tag them appropriately. The bookmarks functionality in Amfora works well enough, but it isn't adequate for my needs. (Not a criticism of Amfora, by the way; Chrome's bookmarks functionality also isn't adequate from my perspective. I have too many bookmarks.)

makeworld-the-better-one:
Definitely some interesting uses there. A PR would be appreciated if the changes aren't too extensive and are documented. Thanks!

mntn-xyz:
Things I didn't add in the PR, but which I considered:
* Automatic substitution of home directory path for ${home} in command path - while this may be useful it's just a nice to have
* Running commands on a specific link from the command bar, like `cmd1:15` to run command 1 on link 15. This could be useful in certain contexts, but honestly it's' not strictly necessary, and I expect people who use this heavily to rebind the keys anyway to sensible mnemonics like "Alt-E" for "email" or similar. In this case they would have to remember that "cmd5" or whatever corresponds to email.

makew0rld:
Fixed by #265. I'll make a new issue for the minor comment I left behind.

PR #265, opened 2021-10-16
Adds support for custom commands

mntn-xyz:
For #262 

Supports up to 10 custom commands, defined in the configuration file. The current or selected URL can be passed to the command by placing ${url} in the command string.

Default hotkeys are Alt-1 through Alt-0 for triggering the command with the selected link URL as a parameter, or Alt-Shift-1 through Alt-Shift-0 for triggering the command with the current page URL as a parameter. This follows the convention set by the copy page URL (C) and copy selected URL (c) defaults.

Custom commands must be available in the $PATH or use absolute paths. Relative paths are not supported, neither are pipes or shell redirections.


makeworld-the-better-one:
Thanks for your PR. I'll try and take a look at this when I have time, but it's not a priority for v1.9.0.

Also note there's now a conflict, let me know if you need any help resolving that.

mntn-xyz:
Thanks, I had not seen the conflict. It is resolved now. Let me know if there's any improvements I need to make when you get a chance to review it.

mntn-xyz:
One of the changes has broken this, need to figure out what happened.

makeworld-the-better-one:
Ok, let me know if you need any help or have questions.

This is a cool feature for sure, but not top of my list, so I'm not sure if it'll make it into the v1.10.0 release. It's gonna be a while before I get there though, so who knows :) Thanks for making contributions to Amfora!

mntn-xyz:
The issue I was encountering seems to be a recently introduced bug, see #281. Once that is fixed, this should be ready.

mntn-xyz:
Todo: use goroutines to prevent deadlock per #281

makeworld-the-better-one:
Please see changes made in #284 (once merged) for a guide on how to fix issues like #281.

makeworld-the-better-one:
@mntn-xyz #284 has been merged and #281 has been closed, so feel free to continue working on this whenever you can. You'll have to merge master back into this PR to get those updates.

mntn-xyz:
Merged up to master, and I made the calls to Error and Info use goroutines as in #284. I think the risk of a race condition here with other parts of the UI is minimal, as it looks like it shouldn't be possible to invoke one of these commands during other operations that also pops up a dialog. The commands are forked when run so long-running commands also won't be an issue. I did some light testing and couldn't intentionally break it.

makeworld-the-better-one:
Sorry I haven't had a lot of time to review this PR. I'm going to add it to the milestone so that I won't forget to come back to it when I make time for Amfora in the future.

command/command.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"os/exec"
+	"strings"
+)
+
+// RunCommand runs `command`, replacing the string "${url}" with `url`.
+func RunCommand(command string, url string) (string, error) {
+	cmdWithURL := strings.ReplaceAll(command, "${url}", url)
+	cmdSplit := strings.SplitN(cmdWithURL, " ", 2)
```
makew0rld, line 11: Using a simple split like this will fail on commands with quoted strings. It might be better to use [go-shellquote](https://github.com/kballard/go-shellquote) or [something else](https://stackoverflow.com/questions/47489745/splitting-a-string-at-space-except-inside-quotation-marks).

makew0rld:
I'll merge this now, I appreciate your work. If you have time to address the issue I mentioned above in a new PR that would be awesome.

PR #271, opened 2021-12-03
Auto redirect if the new URL just adds a trailing slash

awfulcooking:
These redirects are probably necessary, not going anywhere & harmless, let's not prompt for them

e.g. gemini://some-place/~user => gemini://some-place/~user/


makeworld-the-better-one:
The only potential problem I see is that someone could keep redirecting from `gemini://some-place/~user/` to `gemini://some-place/~user//`, etc. But redirects are capped at 5 anyway so it should be fine.

makeworld-the-better-one:
Thanks!

PR #272, opened 2021-12-05
When creating a tab with a URL, show a minimal interstitial page

awfulcooking:
Instead of about:newtab

it's less distracting than about:newtab imo, esp if the network is slow


makeworld-the-better-one:
Nice work! This looks a lot better for sure, thank you.

PR #275, opened 2021-12-08
Makefile: support DESTDIR

aabacchus:
This is the convention and it allows package maintainers to install an
application into an arbitrary rootdir, and does not affect existing
builds.


makeworld-the-better-one:
Merged, thanks! Let me know if you need a new release for this to help with package maintenance.

aabacchus:
no need, thanks :)

PR #277, opened 2021-12-08
🐛 Fix deadlock when an about: URL can't be loaded

awfulcooking:
Caused by display.Error() being made to block for the loading page change.

We must call goURL / handleURL from a goroutine other than the main UI loop

I don't think there are any other problematic calls at the moment

Sorry :)


makeworld-the-better-one:
Good catch, thanks! Let me know if you find anything else.

PR #282, opened 2021-12-11
deps: update to use go1.17

chenrui333:
update golang.org/x/sys to the latest

cc @makeworld-the-better-one

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,30 +1,56 @@
 module github.com/makeworld-the-better-one/amfora
 
-go 1.15
+go 1.17
 
 require (
 	code.rocketnine.space/tslocum/cview v1.5.6-0.20210530175404-7e8817f20bdc
 	github.com/atotto/clipboard v0.1.4
 	github.com/dustin/go-humanize v1.0.0
-	github.com/fsnotify/fsnotify v1.4.9 // indirect
 	github.com/gdamore/tcell/v2 v2.3.3
-	github.com/google/go-cmp v0.5.0 // indirect
 	github.com/makeworld-the-better-one/go-gemini v0.12.1
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/mitchellh/mapstructure v1.3.1 // indirect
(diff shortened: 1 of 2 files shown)


makeworld-the-better-one:
What's the point of this PR? Amfora intends to support Go 1.15+ anyway. Also note that this is failing to build on Go 1.17.

chenrui333:
> What's the point of this PR? Amfora intends to support Go 1.15+ anyway. Also note that this is failing to build on Go 1.17.

Using the latest version of go would help identify the compatibility issue (go 1.16+ has made go modules build as default, and as you can see go1.17 has different layout for `go.mod`).

chenrui333:
I thought I already updated the `go.sum`, let me check on it again.

The `golangci-lint` seems not related to my PR?

makeworld-the-better-one:
> Using the latest version of go would help identify the compatibility issue

What compatibility issue? Amfora has been working fine for all Go versions it's supposed to support.

chenrui333:
> > Using the latest version of go would help identify the compatibility issue
> 
> What compatibility issue? Amfora has been working fine for all Go versions it's supposed to support.

Yeah, it does as of now. 

Let me ask the other way around, what is the challenge of accepting this PR?

makeworld-the-better-one:
> Let me ask the other way around, what is the challenge of accepting this PR?

It doesn't work that way. The author has to prove the value of the PR before it can be accepted. I don't see the point of changing the versions and all these dependencies around. I think it could hurt compatibility for Go versions <1.17 to put `go 1.17` in `go.mod`.

But this is all besides the point because it doesn't even build on Go 1.17:

image

chenrui333:
make sense, we can revisit this along with the go1.18 and later.

makeworld-the-better-one:
You're welcome to submit another PR at some point, but it has to provide a clear advantage over the existing code. Right now I'm happy with the way my module setup is.

chenrui333:
yeah, totally. thanks for the input though! :)

Issue #281, opened 2021-12-11
Error() freezes program when called inside makeNewTab -> t.view.SetInputCapture


mntn-xyz:
Found this because my error messages for #265 started freezing when I merged the latest code from master.

To reproduce, add the following anywhere inside t.view.SetInputCapture (in display/tab.go):

```
Error("Test Error", "Test")
```

For instance, add the above code when the copy key is pressed so it will always have an error when called. Amfora will freeze completely if you try to copy a URL.

I assume this means that the current copy and save errors will freeze the program if they are encountered, but I'm not sure how to trigger those intentionally.

Confirmed that this was introduced in #272.

mntn-xyz:
Drop-in example (display/tab.go):
```
201       err = clipboard.WriteAll(copiedURL.String())                                                                                                                                                      
202       if err != nil {                                                                                                                                                                                   
203         Error("Copy Error", err.Error())                                                                                                                                                                
204         return nil                                                                                                                                                                                      
205       }                                                                                                                                                                                                 
206 +     Error("Test Error", "Test")
207       return nil
208     }
```

awfulcooking:
Hi @mntn-xyz, yes, this is my fault.

I originally made Error() return a channel that could be waited on, but discovered there are a _lot_ of calls that need to wait in display/handlers.go, and chose to make it block instead.

They needed to wait on the modal being dismissed, so that when their parent callers got wind of the error, their activity wouldn't defocus the modal & make it undismissable. In this case, navigating to about:newtab to hide the loading message.

I figured there could be a lot of scenarios like that & so it was logical to make Error() block.

Shortly after it was merged, I found [this freeze](https://github.com/makeworld-the-better-one/amfora/pull/277) which took some time to debug. So I empathize with you.

It seems that CustomCommand() sits in a similar place to [URL()](https://github.com/makeworld-the-better-one/amfora/blob/7edb6c538177d689320a08a38d225e820b4e6163/display/display.go#L553) .. which is called synchronously from UI handlers, but spawns a goroutine to do its navigation so the main loop is free from that point on.

There are a number of other approaches we could take but I'm not sure any of them are simpler or more foolproof...

What do you think?

mntn-xyz:
Thanks for looking into it. Goroutines are probably the right way to do this kind of thing anyway, so overall this change will improve the code. I think it's probably wise to check all the UI code for calls to blocking functions just in case something has slipped by.

[...]

Issue #283, opened 2021-12-11
Download choice modal loses focus when non-text response is received from CLI URL


makeworld-the-better-one:
Run `amfora gemini.thebackupbox.net/radio`. The URL is to a page that returns `audio/mpeg`. The page is loading, then the download choice modal pops up, then the new tab page loads behind it and takes focus, meaning the user can't select a download choice.

makeworld-the-better-one:
Related to #281, if the modal was blocking this wouldn't happen

mntn-xyz:
This also happens if the URL triggers an external handler, for instance "amfora https://example.com" causes the Info dialog to be stuck on screen after the configured browser is opened.

makeworld-the-better-one:
Closed by #284, thanks @awfulcooking!

PR #284, opened 2021-12-12
Modal fixes (#281)

awfulcooking:
Pushing commits to this branch, feel free to test / merge, or wait for more to come in.

### DIFF
--- a/display/tab.go
+++ b/display/tab.go
@@ -178,7 +178,7 @@ func makeNewTab() *tab {
 			currentURL := tabs[curTab].page.URL
 			err := clipboard.WriteAll(currentURL)
 			if err != nil {
-				Error("Copy Error", err.Error())
+				go Error("Copy Error", err.Error())
 				return nil
 			}
 			return nil
@@ -193,14 +193,14 @@ func makeNewTab() *tab {
 			if err != nil {
 				err := clipboard.WriteAll(selectedURL)
 				if err != nil {
-					Error("Copy Error", err.Error())
+					go Error("Copy Error", err.Error())
 					return nil
 				}
 				return nil
 			}
 			err = clipboard.WriteAll(copiedURL.String())
 			if err != nil {
-				Error("Copy Error", err.Error())
+				go Error("Copy Error", err.Error())
 				return nil
 			}
 			return nil
--- a/display/tab.go
+++ b/display/tab.go
@@ -156,12 +156,12 @@ func makeNewTab() *tab {
 			if t.hasContent() {
 				savePath, err := downloadPage(t.page)
 				if err != nil {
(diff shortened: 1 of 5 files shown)


makeworld-the-better-one:
Looking good, thanks for this! Let me know when you think it's done.

display/tab.go
```diff
@@ -156,12 +156,12 @@ func makeNewTab() *tab {
 			if t.hasContent() {
 				savePath, err := downloadPage(t.page)
 				if err != nil {
-					Error("Download Error", fmt.Sprintf("Error saving page content: %v", err))
+					go Error("Download Error", fmt.Sprintf("Error saving page content: %v", err))
```
makeworld-the-better-one, line 159: My problem with this, and all the other `go <modal>` calls in this file is that it opens up the possibility of other things happening before the modal is dismissed. I'm not sure what the solution is though. Maybe all the code in `t.view.SetInputCapture` can be put in a goroutine (with a mutex that blocks multiple versions of it running) that blocks? I'm not sure.

awfulcooking: This might be a good situation to ask about in #go-nuts on Libera. I wonder if they'd have any advice?

awfulcooking: Bear in mind the Info() and Error() functions previously just didn't block, but would return as soon as the modal had opened. Using `go Info()` should give similar behaviour to before, the only difference being that now the parent function may return before the modal has even rendered.

But if anything happened immediately *after* it the modals rendered, it would have shown UI breakage before #272, I think. And so we could expect this just to be a return to the previous, non-deadlocking behaviour

awfulcooking: Fun that we're addressing deadlocking and not-waiting-enough code in one PR :-)
As two sides of one coin.

awfulcooking:
I think we should merge this, test it, and release after a few days if all seems well. Then improve from there!

makeworld-the-better-one:
I'd rather test before merging, but in general this seems like a good fix for this issue, thanks for your work. As for a release, I gotta see, because I already have a new feature in `master` (#263), so to make a v1.9.3 release for #283 would require reverting the feature and re-adding it after release. #283 is not a show-stopper bug, so I'm not sure if it merits its own release or not.

Also please see my reply above.

makeworld-the-better-one:
As for testing, it would probably be best to induce errors, like was done in #281, and make sure there are no UI issues.

makeworld-the-better-one:
Going to merge this now, having done some testing myself. Thanks for making this patch. I don't intend to make a release for this, but it will be in v1.10.0 of course.

If you have an idea about my [comment](https://github.com/makeworld-the-better-one/amfora/pull/284#discussion_r767899077) above that would be appreciated.

Issue #182, opened 2021-02-06
Table of Contents


ThomasAdam:
Hey @makeworld-the-better-one 

As others have echoed many times, thanks for this client -- it's certainly one of the better terminal ones!

I was looking at the TODO list, and noticed you had an item on there which I'm keen to work on, assuming it's not already being implemented elsewhere (in which case I'll look at some other feature to code), and that's adding a ToC (Table of Contents).

In adding this, I have some questions before I start submitting PRs for this.  The first one is to do with the parser (in `renderer.go`).  At the moment, the parser is stateless in that it's rendering the page line-by-line, and building up a display output which can be rendered.   I'm actually wondering if we should instead have the concept of a `Parser`, perhaps something like:

```go
type Parser struct {
	scanner           *bufio.Scanner
	Lines             []Line
	isPre             bool
	tableofcontents   []*Heading // see toc.go
}
```

Where we have a `bufio.Scanner` representing the raw gemini lines, and that `Lines` is out struct of parsed content.

From there, we can introduce further types that know how to handle:

* Headings
* Links
* Quotes

Most of this is yak-shaving to then be able to build up a ToC though, which is my next point.  As  I see it, a ToC would have to represent two things:

1. An overview of the document, with backlinks to the rendered document so that it's scrolled at the appropriate point (think HTML anchors `foo.html#whatever`)
2. Be displayed somewhere convenient on the UI

Did you have anything in mind with regards to point 2?  I was imagining a sidebar on either the left or right of the screen which can be toggled on and off, perhaps?

Point 1 is just a variation on a theme about getting the `Parser` into a shape where adding in these ephemeral states which fall outside of pure gemini-rendering become easier.

Let me know your thoughts, and I'll happily crack on with this.

Thanks again!
Thomas

makeworld-the-better-one:
Just a heads up, I edited your parser code example to make it more correct and readable, feel free to change it if I made a mistake.

Thanks for the compliment, and I'm happy to hear you're interested in working on this feature!

>  I'm actually wondering if we should instead have the concept of a `Parser`

[...]

PR #293, opened 2022-01-26
Add top level 1 heading as bookmark title suggestion

### LINKED ISSUE #267: When bookmarking a page, suggest a title based on the first level 1 heading
I usually just retype a short version of the heading, what do other people do?

amohamed11:
Added basic top level 1 `#` heading extraction to suggest it as title for bookmark. I'm a regex newbie, but this was the best regex I could come up with that handled most cases that I tested for. Any regex wizardry advice is welcome. 

Closes #267

### DIFF
--- a/display/bookmarks.go
+++ b/display/bookmarks.go
@@ -2,6 +2,8 @@ package display
 
 import (
 	"fmt"
+	"regexp"
+	"strings"
 
 	"code.rocketnine.space/tslocum/cview"
 	"github.com/gdamore/tcell/v2"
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
Thanks for your PR, this looks simple and nice! It may be replaced once #182 is implemented, but that's not happening any time soon.

However, I think your regex is faulty. I played around and came up with `(?m)^#[^#][\t ]*[^\s].*$` which I think is better. Anything can be in a header, but your regex disallows Unicode symbols. It also matches across newlines due to use of `\s`.

If you didn't know already, you can test and play with your regex at https://regex101.com/. It's a great site and very handy for understanding regex and creating your own.

If you don't spot any issues with my regex, you can add it to this PR and I'll merge it after testing.

amohamed11:
@makeworld-the-better-one ah good catch on the unicode, completely forgot to test for that. I tested your regex on a few examples, and it seemes to work well. I've updated the regex and amended the commit to keep things small. Let me know if there is anything else.

makeworld-the-better-one:
Thanks for this, works great! Will merge now. In the future, I prefer contributors avoid force-pushes / amending, so what was changed is clear. I will squash the commits after anyway.

PR #298, opened 2022-02-21
add rosé pine variants

mvllow:
adds the three rosé pine variants 😌 cheers!


makeworld-the-better-one:
Thanks for adding this! You also need to update the theme README with some information and a photo, see https://github.com/makeworld-the-better-one/amfora/pull/251 or any other theme PR as an example.

mvllow:
Ah apologies for missing that, may be a few days but I'll do that asap 😌

makeworld-the-better-one:
Thank you! Please subscribe to https://github.com/makeworld-the-better-one/amfora/issues/149 so you can stay aware of themeing updates.

Issue #296, opened 2022-02-13
Consider shipping release tarball with vendored dependencies


illiliti:
For example, [dnscrypt-proxy](https://github.com/DNSCrypt/dnscrypt-proxy/tree/master/vendor) does this by simply maintaining vendor directory. You can also setup CI to vendor dependencies into release tarball without needing to keep vendor directory within repository. Here is the command that can be used for this: `go mod vendor`

With vendored dependencies, i can build amfora from source without internet connection by using this command: `go build -mod=vendor`

makeworld-the-better-one:
Good idea, thanks. I'll add this for the next release.

makeworld-the-better-one:
PR #299 is one attempt at adding this.

aabacchus:
@makeworld-the-better-one has there been any progress on this? Thanks

makeworld-the-better-one:
Nope I haven't made progress on this. It's still something I'd like to do but not a high priority for me. I made a comment in #299, so if someone made a PR that addresses that comment, I might be able to merge it in.

makew0rld:
Closing this as there will be few feature Amfora releases. However I see the value and so I have manually added a vendored source code tarball to the latest release.

PR #299, opened 2022-02-26
Generate release tarball in GitHub Actions

CyberTailor:


makeworld-the-better-one:
I would prefer if `.goreleaser.yml` could remain unchanged, to prevent vendoring if I run goreleaser on my local machine. But I will revaluate this PR more closely when I come around to #296 myself. Thanks for making it.

makew0rld:
Closing manually due to inactivity and lack of Amfora development.

Issue #301, opened 2022-03-26
feature request: add confirmation promps when visiting non-gemini URLs


emily-is-my-username:
Hi!
When using amfora, I keep accidentally opening browser windows by selecting http(s) urls that are not clearly marked as such.
I looked for a ways to verify non-gemini URLs before visiting them, but none of the options I found were really convenient for me:
- enable `show_link` (I don't like the visual clutter)
- use tab to select the links and verify them in the URL bar (slow compared to using link numbers)
- disabling the URLs scheme entirely (even slower because then I have to copy-paste to other applications)
- using an external handler to show the prompt (complicated to set up) 

IMO, a (y/n) confirmation prompt (similar to redirects) on visiting URLs with a configurable set of schemes would be a good tradeoff between visual clutter, control and convenience.

EDIT:
Sorry, I just noticed that the `underline` option exists to mark non-gemini-URLs.
I will still leave this open in case it is wanted. I personally like being able to quickly review the URL before opening it.

makeworld-the-better-one:
Hi, thanks for your interest in Amfora.

Personally I differentiate between HTTP and Gemini URLs using themeing. Here's an example just on the new tab page of my Amfora setup:

![image](https://user-images.githubusercontent.com/25111343/160305083-5174939a-34d9-4030-898d-3fbc295a8e36.png)

This theme used to be the Amfora default before v1.9.0. Copied from the [release notes](https://github.com/makeworld-the-better-one/amfora/releases/tag/v1.9.0):

> The default theme of Amfora was changed, now the user's default terminal colors are used. See #181 for details. If you want the original Amfora theme back, copy [this file](https://github.com/makeworld-the-better-one/amfora/blob/master/contrib/themes/amfora.toml)  into your config, replacing the `[theme]` section.

Using different colours is very clear to me and does the job. Thanks for adding a PR, and I think the idea of scheme-based prompts is not bad. But if your issue is "accidentally opening browser windows by selecting http(s) urls", maybe it's unnecessary and themeing is a good enough fix? Let me know what you think.

emily-is-my-username:
> Personally I differentiate between HTTP and Gemini URLs using themeing.

Thank you for the reminder. That was another oversight on my part, I guess I subconsciously assumed those were already-visited URLs (even though I know amfora does not track those) without thinking twice about it.

[...]

PR #302, opened 2022-03-26
add confirmation prompts for url schemes

emily-is-my-username:
This commit adds a confirmation prompt before following urls with
with certain schemes.
Wether or not a scheme requires confirmation can be configured in the
new `[url-promps]` section.

For #301


makeworld-the-better-one:
Thanks for this, it's handy. Merged.

PR #304, opened 2022-04-05
Add half_pgup and half_pgdn bindings

### LINKED ISSUE #303: Binding Half Page Down
Currently, `bind_pgdn` seems to scoll about 3/4 of a page. I find it difficult to use page up and down because I'm looking to continue reading from the wrong part of the screen. I guess this is learnt behaviour from years of using Vim and Less which both scroll half a page with page up and down (though that might be config on my side).

I would really like some new bindings that will allow me to scroll half a page at a time.

Thanks.

Maugrift:
Adds `half_pgup` and `half_pgdn` bindings (`CmdHalfPgup` and `CmdHalfPgdn`), which scroll by half a page, like vim and less. These are unbound by default, but are listed in the default config file and have a help entry.

This was implemented by refactoring `pageUp` and `pageDown` into a generic `relativeScroll` method called by all page up/down functions. There's a change that the float conversions might not yield exactly the same scroll distance, but it shouldn't be off by more than a line if so.

Fixes #303

### DIFF
--- a/config/keybindings.go
+++ b/config/keybindings.go
@@ -49,6 +49,8 @@ const (
 	CmdMoveRight
 	CmdPgup
 	CmdPgdn
+	CmdHalfPgup
+	CmdHalfPgdn
 	CmdNewTab
 	CmdCloseTab
 	CmdNextTab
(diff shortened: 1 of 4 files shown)


makeworld-the-better-one:
Thanks for your contribution! I've opted to just change the behaviour of the original paging bindings, see #303.

PR #324, opened 2022-09-28
URL no longer escaped and unescaped during normalization

### LINKED ISSUE #322: don't url-decode slash or plus in redirect URLs
If amfora receives a response that looks like this:
```
30 gemini://example.org/some%2Fpath/to%2Ba%2Bpage/
```
It decodes this and tries to access the page at
```
gemini://example.org/some/path/to+a+page/
```
This is a different url to the one given in the redirect -- the `/` character is a path separator, but `%2F` is not. Amfora should not decode these characters when handling redirects.

autumnull:
Fixes #322


makeworld-the-better-one:
Thanks for this fix!

Issue #326, opened 2022-11-26
localhost address


FiskFan1999:
Currently, if you type "localhost" into the bar, it defaults to a geminispace.info search for that term, as opposed to a direct link with scheme gemini:// if you type a URL such as gemini.circumlunar.space (this happens to me a lot because I enjoy using amfora to test my gemini capsule locally). This doesn't happen if you call `amfora localhost`, because the address in argv is always treated as direct URL. I'm interested in your thoughts on if localhost should also result in a direct link instead of a search. Thank you

makeworld-the-better-one:
Yeah that's fair, I think it makes sense for `localhost` to go to `gemini://localhost/` in the bottom bar. I'm not working on Amfora much these days but would accept a PR for this.

FiskFan1999:
Thanks. Will get on this. Thanks for making amfora by the way.

makew0rld:
Fixed in #327

PR #327, opened 2022-11-26
localhost is direct URL now instead of geminispace.info/search?localhost (#326)

FiskFan1999:
Small logic change. Supports typing `localhost`, `localhost:port`, and `localhost/path/to/file` (and `localhost:port/path/to/file`).

Let me know if this is missing anything or if I messed something else up. thanks


FiskFan1999:
based on #326

PR #328, opened 2022-11-30
Allow following all links with just the number keys

codesoap:
I find it mildly annoying to have to type `<spc> + <link-number> + <return>` for links with numbers greater than ten. I thought about how the number of key presses could be reduced and came up with the idea to use the (leading) zero as a special character. Link-numbers larger than nine would get one leading zero, ones larger than 99 would get two, ones larger than 999 would get three and so forth. A picture can probably explain it better:

demo

With this change links 1-9 would work as before, for 10 the user would have to type `010` instead of `0`. For 11-99 the user would just need to press three keys instead of four, which is the improvement this pull request is about. I visit a lot of pages that contain more than ten but less than 100 links, so this change would often be useful to me.

PS: This is not meant to be a polished pull request. I just wanted you to be able to try my idea. If you like it, I can continue by adapting the documentation, ...

PPS: I also thought about other ways of reducing key presses for links. One would be to allow some letters in "link selectors", so that ~30-50 links could be followed with a single key press, but I figured it would be too much of a change for the users.

### DIFF
--- a/display/tab.go
+++ b/display/tab.go
@@ -231,13 +231,25 @@ func makeNewTab() *tab {
 			}
 			return nil
 		}
-		// Number key: 1-9, 0, LINK1-LINK10
+		// Number key: 1-9, 0
 		if cmd >= config.CmdLink1 && cmd <= config.CmdLink0 {
-			if int(cmd) <= len(t.page.Links) {
-				// It's a valid link number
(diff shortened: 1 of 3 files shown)


makew0rld:
Thanks for your PR. I'm happy with the current system however, and don't want to indent links further in rendering.

PR #336, opened 2023-08-29
Update go dependencies and require go 1.17

grisu48:
Require go 1.17 as needed by golang.org/x/sys/unix/syscall and update all go dependencies to current versions, which are not pinned.

Proposed fix for https://github.com/makew0rld/amfora/issues/335


makew0rld:
Thanks for this fix. Please don't update `cview` as it may cause UI breaking changes. It's pinned to a specific commit on purpose. Please also check semantic versioning, for example `termenv` may have breaking changes as well according to the semantic versioning spec. Otherwise I will be happy to merge this.

grisu48:
Thanks for the review! I'll have a look and update the PR accordingly.

grisu48:
@makew0rld do those change address your concerns now?

makew0rld:
Thanks!

grisu48:
Thanks to you :slightly_smiling_face:

PR #349, opened 2024-09-21
Certificates support

smoorg:
This change introduces breaking change of calling cmd to `openssl`.

That way I was able to achieve support for creating certificates and
keys straight from amfora.

The intention was, once user gets invalid certificate code, amfora acts
proactively and proposes to create the cert for that page on its own.

Concerns:
1. This might break Windows build. I do not know much about how openssl is
implemented on its part. I should probably prepare some additional patch
or exclude it from that implementation.
2. We need openssl dependency on package definition end.


makew0rld:
Sorry but I am not interested in depending on OpenSSL, this should be possible to do fully with the high quality API the Go stdlib provides.

Also please note that Amfora is in maintenance mode and so I can't guarantee such a feature will ever be merged.

smoorg:
I understand. I'll leave it on my fork then. I managed to make it work with `crypto` package so thanks for the feedback.

PR #351, opened 2024-10-23
Use preformatted_text color for rendering plain text files

seifferth:
Using a color from the theme should make the rendering of text files be more in line with the colors chosen for styled content. Furthermore, allowing some form of user override for the color of plain text files allows users to individually fix bad color combinations. Such as in my case, where the combination of a solarized terminal theme and the default color chosen by amfora causes plain text files to be rendered as white text on an almost equally bright background.

### DIFF
--- a/renderer/renderer.go
+++ b/renderer/renderer.go
@@ -49,7 +49,7 @@ func RenderPlainText(s string) string {
 	// It used to add a left margin, now this is done elsewhere.
 	// The function is kept for convenience and in case rendering
 	// is needed in the future.
-	return cview.Escape(s)
+	return fmt.Sprintf("[%s]", config.GetColorString("preformatted_text")) + cview.Escape(s)
 }
 
 // wrapLine wraps a line to the provided width, and adds the provided prefix and suffix to each wrapped line.


makew0rld:
> Such as in my case, where the combination of a solarized terminal theme and the default color chosen by amfora causes plain text files to be rendered as white text on an almost equally bright background.

Can you explain more about this issue? Why doesn't it occur when browsing gemtext files?

seifferth:
> Can you explain more about this issue? Why doesn't it occur when
> browsing gemtext files?

Basically I am using xfce4-terminal with the solarized light color
scheme. While this actually looks quite good, it usually combines
horribly with applications that try to do more fancy color handling
themselves. Especially since the color scheme only redefines 16 colors
while those applications often use 256, which often causes a certain
misalignment of color schemes.

As a solution, I usually just redefine the color scheme in the
config file. That's also what I did in amfora. The default settings
for text/gemini would also render white text on an almost white
background. It's just that I had no way to redefine the color used
for text/plain.

Furthermore, I believe that there are specific terminal color codes for
"foreground" and "background". While these may sometimes be the same as
"white" and "black", they are not guaranteed to be. In the particular
case of light color schemes, they are usually even inverted. So as a
general rule, I believe using "default foreground" rather than "white"
is the safer default.

makew0rld:
Makes sense, thanks.

PR #360, opened 2025-05-27
Fix build error resulting from cview's move to codeberg

h14h:
Recently tried to `brew install --HEAD amfora` to get the latest and greatest, only to encounter a build failure.

Turns out this was a result of cview migrating their repo to codeberg:
https://codeberg.org/tslocum/cview/commit/242e7c1f1b61a4b3722a1afb45ca1165aefa9a59

Added a replace directive to point to the new location, and `make` now results in a successful build.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,3 +56,5 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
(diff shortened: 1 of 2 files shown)


makew0rld:
Thanks for finding this and maintaining the same commit reference.

stefanb:
It would be nice to get a new release, one that is buildable.

Detected in
* https://github.com/Homebrew/homebrew-core/pull/226636

PR #364, opened 2025-10-30
Update dependencies

grisu48:
Update all dependencies to the most recent version.

This also fixes two low impact vulnerabilities in golang.org/x/net/html, which result in [calls to Parse executing unexpectedly slow](go.dev/issue/75682). The vulnerability is probably not really relevant for this project though.

I've compiled and tested the binaries with those dependencies and on my system I could not observe anything out of the ordinary.


grisu48:
Closing in favor or https://github.com/makew0rld/amfora/pull/365. Thank you!

